PR #4, opened 2019-07-10
OPENEUROPA-2056: Better translation form elements.

upchuk:


tests/features/local_translations.feature
```diff
@@ -21,9 +23,17 @@ Feature: Local translations
     When I click "Translate locally" in the "Bulgarian" row
     Then the translation form element for the "Title" field should contain "My title"
     And the translation form element for the "Body" field should contain "My body"
+    And the translation form element for the "Link Uri (1)" field should contain "https://example.com"
+    And the translation form element for the "Link Title (1)" field should contain "Example"
+    And the translation form element for the "Link Uri (2)" field should contain "/node"
+    And the translation form element for the "Link Title (2)" field should contain "Node"
```
imanoleguskiza: I kinda agree with @dxvargas here, why not create a different step "Given a translatable node with the "My title" title and "My body" body and the links:" that receives a table of link-title pairs? It would make it more readable.

tests/features/local_translations.feature
```diff
@@ -41,8 +51,25 @@ Feature: Local translations
...
+    When I fill in the translation form element for the "Title" field with "Bulgarian title 2"
+    And I press "Save and complete translation"
+    Then I should see "Translations of Bulgarian title 2"
+    When I click "Bulgarian title 2" in the "Bulgarian" row
+    Then I should see "Bulgarian title 2"
+    And I should see "Bulgarian body"
+
   Scenario: Preview translations
-    Given a translatable node with the "My title" title and "My body" body
+    Given a translatable node with the "My title" title and "My body" body and multiple links
```
imanoleguskiza: Related to the comment above, here we add links when they are not needed? If we had separate steps it would be clearer in my opinion

PR #7, opened 2019-08-02
OPENEUROPA-2105: Poetry client and mock setup.

upchuk:


modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/PoetryMock.php
```diff
@@ -0,0 +1,56 @@
...
+   * @param string $password
+   *   The poetry service password.
+   * @param string $message
+   *   The message.
+   *
+   * @return string
+   *   The XML string response.
+   */
+  public function requestService(string $username, string $password, string $message): string {
+    if ($username !== 'admin' || $password !== 'admin') {
```
imanoleguskiza: I think we should probably test this as well on the PoetryMockTest test class.

modules/oe_translation_poetry/modules/oe_translation_poetry_mock/oe_translation_poetry_mock.routing.yml
```diff
@@ -0,0 +1,20 @@
+oe_translation_poetry_mock.server:
+  path: '/poetry-mock/server'
+  defaults:
+    _controller: 'Drupal\oe_translation_poetry_mock\Controller\PoetryMockController::server'
+  requirements:
+    _permission: 'access content'
+
+oe_translation_poetry_mock.server_types:
```
imanoleguskiza: What is the purpose of this route?

modules/oe_translation_poetry/tests/Functional/PoetryMockTest.php
```diff
@@ -0,0 +1,109 @@
...
+      $this->assertEqual($setting, $settings[$name]);
+    }
+
+    // Create a test message and check that Mocked responses are returned.
+    /** @var \EC\Poetry\Messages\Requests\CreateTranslationRequest $message */
+    $message = $poetry->get('request.create_translation_request');
+
+    $test_id = 1;
+    $id = $message->getIdentifier();
+    $id->setNumber('40012');
```
imanoleguskiza: Usually on production we only set the numero once we have received one from poetry, on the very first request you should send a sequence identifier and receive the numero from poetry.

runner.yml.dist
```diff
@@ -39,6 +39,27 @@ commands:
...
+    - { task: "process-php", type: "write", config: "drupal.settings", source: "${drupal.root}/sites/default/default.settings.php", destination: "${drupal.root}/sites/default/settings.testing.php", override: true }
+    - task: "append"
+      file: "build/sites/default/settings.testing.php"
+      text: |
+        $settings["poetry.identifier.code"] = getenv('POETRY_IDENTIFIER_CODE');
+        $settings["poetry.identifier.sequence"] = getenv('POETRY_IDENTIFIER_SEQUENCE');
+        $settings["poetry.service.username"] = getenv('POETRY_SERVICE_USERNAME');
+        $settings["poetry.service.password"] = getenv('POETRY_SERVICE_PASSWORD');
+        $settings["poetry.notification.username"] = getenv('POETRY_NOTIFICATION_USERNAME');
+        $settings["poetry.notification.password"] = getenv('POETRY_NOTIFICATION_PASSWORD');
```
imanoleguskiza: Why do we not set the mock as the default poetry endpoint in the tests? I know we set it in the setup of the test class but do you foresee at any point using anything but a mock during tests?

modules/oe_translation_poetry/modules/oe_translation_poetry_mock/oe_translation_poetry_mock.routing.yml
upchuk: Unclear at the moment, it exists in D7. There is a todo to remove it.

runner.yml.dist
upchuk: This to create a .testing.settings file that is used in the test as the default one is not included.

PR #8, opened 2019-08-07
OPENEUROPA-2107: Create Poetry translation requests

upchuk:


modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -41,7 +94,167 @@ public function __construct(ConfigFactoryInterface $configFactory, LoggerChannel
...
+    // If the incremented part is 100, we need to scrap the the global number
+    // and request a new one. The maximum can be 99.
+    if ($part === 100) {
+      $identifier->setSequence(Settings::get('poetry.identifier.sequence'));
+      return $identifier;
+    }
+
+    $identifier->setNumber($number);
+
+    if ($part > -1) {
```
imanoleguskiza: If I follow the code correctly, if the getLastPartForNumber ever returns a -1 we will set the part to be 0 (as defined in the constructor). However, if we ever get a -1 I believe we should behave differently since this would mean we have lost track of the part for the particular number we are checking for. In this case we should either throw and exception or request a new number right?

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -0,0 +1,410 @@
...
+        ->setNickname($form_state->getValue('details')['organisation'][$name]);
+    }
+
+    // Build the return endpoint information.
+    // @todo update once we implemented the notification handling.
+    $return = $message->withReturnAddress();
+    $return->setUser('test');
+    $return->setPassword('test');
+    // The notification endpoint WSDL.
+    $return->setAddress(Url::fromRoute('<front>')->setAbsolute()->toString());
```
dxvargas: This, user and password could already be set using config values.

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -0,0 +1,410 @@
...
+
+    // Update all the jobs with the resulting identifier.
+    foreach ($jobs as $job) {
+      $job->set('poetry_request_id', $identifier_values);
+    }
+
+    // Submit all the job entities. This will also save them.
+    foreach ($jobs as $job) {
+      $job->submitted();
+    }
```
dxvargas: There is no need to iterate on $jobs twice.

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -0,0 +1,410 @@
...
+    ];
+
+    $form['details']['date'] = [
+      '#type' => 'date',
+      '#title' => $this->t('Requested delivery date'),
+      '#required' => TRUE,
+    ];
+
+    $default_contact = $translator_settings['contact'] ?? [];
+    $form['details']['contact'] = [
```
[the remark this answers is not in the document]
upchuk: Made them required on the request form itself, not the translator form. It's not technically needed to set global values.

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -0,0 +1,410 @@
...
+      // identifier number with the new number that came for future requests.
+      if ($identifier->getSequence()) {
+        $this->poetry->setGlobalIdentifierNumber($response->getIdentifier()->getNumber());
+      }
+
+      $this->redirectBack($form_state);
+      $this->messenger->addStatus($this->t('The request has been sent to DGT.'));
+    }
+    catch (\Exception $exception) {
+      $this->messenger->addError($this->t('There was a error making the request to DGT.'));
```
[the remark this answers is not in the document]
upchuk: Done.

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -0,0 +1,410 @@
...
+      $this->handlePoetryResponse($response);
+
+      // If we request a new number by setting a sequence, update the global
+      // identifier number with the new number that came for future requests.
+      if ($identifier->getSequence()) {
+        $this->poetry->setGlobalIdentifierNumber($response->getIdentifier()->getNumber());
+      }
+
+      $this->redirectBack($form_state);
+      $this->messenger->addStatus($this->t('The request has been sent to DGT.'));
```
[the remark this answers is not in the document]
upchuk: OK, I think I solved this inside `handlePoetryResponse()`. Check please.

modules/oe_translation_poetry/src/Poetry.php
upchuk: How about now? If we get -1, we set the sequence to request a new number.

modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
upchuk: Will handle this entire thing in the notifications ticket.

modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -41,7 +94,167 @@ public function __construct(ConfigFactoryInterface $configFactory, LoggerChannel
...
+   */
+  protected function getLastIdentifierForContent(ContentEntityInterface $entity): ?Identifier {
+    $query = $this->database->select('tmgmt_job', 'job');
+    $query->join('tmgmt_job_item', 'job_item', 'job.tjid = job_item.tjid');
+    $query->fields('job');
+    $query->condition('job_item.item_id', $entity->id());
+    // Do not include unprocessed Jobs. These are the ones which have not been
+    // ever sent to Poetry.
+    $query->condition('job.state', Job::STATE_UNPROCESSED, '!=');
+    $query->orderBy('job.poetry_request_id__version', 'DESC');
```
dxvargas: No need to return the full array of jobs.
```$query->range(0, 1);```

modules/oe_translation_poetry/src/Poetry.php
imanoleguskiza: I think this is a solution that works from our side but will not be very liked in the DGT side. In any case, I don't think we have any better solution for now.

PR #13, opened 2019-08-09
OPENEUROPA-2157: Add configuration to translator.

imanoleguskiza:
### Description

Add configuration to translator so we can have default values for the Poetry requests..

### Change log

- Added: Add configuration to translator.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_translation_poetry/config/install/tmgmt.translator.poetry.yml
```diff
@@ -1,13 +0,0 @@
-langcode: en
```
upchuk: Why we not shpping with a default translator?

modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -21,11 +22,13 @@ class Poetry extends PoetryLibrary {
...
+   * @param \Drupal\tmgmt\TranslatorManager $translatorManager
+   *   The translator plugin manager.
    */
-  public function __construct(ConfigFactoryInterface $configFactory, LoggerChannelFactoryInterface $loggerChannelFactory) {
-
+  public function __construct(ConfigFactoryInterface $configFactory, LoggerChannelFactoryInterface $loggerChannelFactory, TranslatorManager $translatorManager) {
+    $poetry_translator_definition = $translatorManager->getDefinition('poetry');
     $values = [
-      'identifier.code' => Settings::get('poetry.identifier.code'),
+      'identifier.code' => $poetry_translator_definition['default_settings']['identifier'],
```
upchuk: This is not correct. You are taking the hardcoded data from the the plugin definition. What happens if the config in the translator changes? It wont change the plkugin definition. I think we might not even need to set defaults in the plugin definition. Not sure why they are there but for sure this is not how they are used. Perhaps as the defalt value in the config form.

modules/oe_translation_poetry/config/schema/oe_translation_poetry.schema.yml
```diff
@@ -28,3 +27,44 @@ field.value.poetry_request_id:
     product:
       type: string
       label: Product
+tmgmt.translator.settings.poetry:
+  type: tmgmt.translator_base
+  mapping:
+    identifier:
```
upchuk: `identifier` is the entire thing. This is the identifier_code

modules/oe_translation_poetry/config/schema/oe_translation_poetry.schema.yml
```diff
@@ -28,3 +27,44 @@ field.value.poetry_request_id:
     product:
       type: string
       label: Product
+tmgmt.translator.settings.poetry:
+  type: tmgmt.translator_base
+  mapping:
```
upchuk: Should these config values be under the `settings` key of the translator config by the way?

PR #15, opened 2019-08-19
OPENEUROPA-2106: Handle poetry notifications

upchuk:


modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
```diff
@@ -344,4 +652,22 @@ protected function assertJobsPoetryRequestIdValues(array $jobs, array $values):
...
+  /**
+   * Groups the jobs by their target language.
+   *
+   * @param \Drupal\tmgmt\Entity\JobInterface[] $jobs
+   *   The jobs.
+   *
+   * @return \Drupal\tmgmt\Entity\JobInterface[]
+   *   The jobs.
+   */
+  protected function groupJobsByLanguage(array $jobs): array {
```
imanoleguskiza: Nickpicking here, but the method should be named indexJobsByLanguage since we are not grouping multiple jobs under any given language but using the language of the jobs as the keys for the array.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
```diff
@@ -275,6 +287,302 @@ public function testUnprocessedJobHandling(): void {
...
+  /**
+   * Tests the handling of Poetry status notifications.
+   */
+  public function testStatusNotifications(): void {
+    $this->prepareRequestedJobs([
+      'title' => 'My node title',
+      'field_oe_demo_translatable_body' => 'My node body',
+    ], ['fr', 'de', 'it']);
+
+    $identifier_info = [
```
imanoleguskiza: We could probably define a "defaultIdentifier" parameter in the setup, but not required.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
```diff
@@ -275,6 +287,302 @@ public function testUnprocessedJobHandling(): void {
...
+          'accepted_date' => '30/09/2019 23:59',
+        ],
+      ]);
+
+    $this->performNotification($status_notification);
+
+    // Send the translations for each job.
+    $jobs = $this->groupJobsByLanguage($this->jobStorage->loadMultiple());
+
+    foreach ($jobs as $job) {
```
imanoleguskiza: We are only testing a single translation so is there need to do this foreach loops? The whole loop is then done again on #434 so removing it would simplify the whole test method a lot.

modules/oe_translation_poetry/src/EventSubscriber/PoetryNotificationSubscriber.php
```diff
@@ -0,0 +1,325 @@
...
+      /** @var \Drupal\tmgmt\JobInterface $job */
+      $job = $jobs[$language] ?? NULL;
+      if (!$job) {
+        $this->logger->error('There is a missing job for the language @lang and request ID @id', ['@lang' => $language, '@id' => $identifier->getFormattedIdentifier()]);
+        continue;
+      }
+
+      // Set the date that comes from Poetry.
+      $this->updateJobDate($job, $target);
+    }
```
dxvargas: For me it would be more logic to handle the job dates inside the block below:
```
if ($job_status_code === 'ONG') {...}
```

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
```diff
@@ -275,6 +287,302 @@ public function testUnprocessedJobHandling(): void {
...
+      $items = $job->getItems();
+      $item = reset($items);
+      $data = $this->container->get('tmgmt.data')->filterTranslatable($item->getData());
+      foreach ($data as $field => &$info) {
+        $info['#text'] .= ' - ' . $job->getTargetLangcode();
+      }
+
+      $generator = $this->container->get('oe_translation_poetry_mock.fixture_generator');
```
dxvargas: There is already `$this->fixtureGenerator`.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
```diff
@@ -275,6 +287,302 @@ public function testUnprocessedJobHandling(): void {
     $this->assertJobsPoetryRequestIdValues($this->jobStorage->loadMultiple(), $expected_poetry_request_id);
   }
 
+  /**
+   * Tests the handling of Poetry status notifications.
+   */
+  public function testStatusNotifications(): void {
```
dxvargas: I would put this and next tests in a different place e.g. `PoetryTranslationNotificationTest`.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
upchuk: OK, will do.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
upchuk: OK will do.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
upchuk: Will do.

modules/oe_translation_poetry/modules/oe_translation_poetry_mock/fixtures/translation_template.xml
```diff
@@ -0,0 +1,25 @@
...
+
+    <title>Job ID @job_id</title>
+  </head>
+  <body>
+    <div class="asset" id="item-@item_id">
+      <!--
+  label="Title"
+  context="[@item_id][title][0][value]"
+  -->
+      <div class="atom" id="bMV1bdGl0bGVdWzBdW3ZhbHVl">@title][0][value</div>
```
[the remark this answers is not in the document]
upchuk: This I fixed in my current ticket

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
upchuk: OK will do.

modules/oe_translation_poetry/tests/Functional/PoetryTranslationRequestTest.php
upchuk: Yes, true, will do.

PR #17, opened 2019-08-20
OPENEUROPA-2187: Allow users to clean unprocessed poetry jobs.

imanoleguskiza:
### Description

If not dealt with inmediatly some poetry jobs may be left unprocessed. Allow users to deal with this jobs.
At the same time, if there are still jobs in the job queue, allow users to finish processing them.
### Change log

- Added: Links to content translation table to remove unprocessed jobs.
- Changed: Allow users to finish pending requests.
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -184,6 +184,25 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
   }
 
+  /**
+   * Returns the title of the form page.
+   *
+   * @return string
+   *   The title for the form page.
+   */
+  public function getPageTitle(): string {
+    $title = 'Send request to DG Translation';
```
upchuk: This needs to be translatable I think. Make sure it's in t(). And not `t($var)`. You can use placeholders for the languages.

PR #18, opened 2019-08-20
OPENEUROPA-2106: Rework the Poetry client service.

upchuk:


modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -90,29 +139,33 @@ public function __construct(array $settings, ConfigFactoryInterface $configFacto
       'service.password' => Settings::get('poetry.service.password'),
       'notification.username' => Settings::get('poetry.notification.username'),
       'notification.password' => Settings::get('poetry.notification.password'),
-      'logger' => $loggerChannel,
+      'notification.endpoint' => Url::fromRoute('oe_translation_poetry.notifications')->setAbsolute()->toString(),
```
dxvargas: Missing route definition.

modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -90,29 +139,33 @@ public function __construct(array $settings, ConfigFactoryInterface $configFacto
...
+  /**
+   * Delegates to the Poetry library all calls made to this service.
+   *
+   * {@inheritdoc}
+   */
+  public function __call($name, $arguments) {
+    $this->initialize();
+    if (method_exists($this->poetryClient, $name)) {
+      return call_user_func_array([$this->poetryClient, $name], $arguments);
+    }
```
dxvargas: And if method is not defined...? Shouldn't it throw an error?

modules/oe_translation_poetry/src/Poetry.php
upchuk: True, fixed.

modules/oe_translation_poetry/src/Poetry.php
upchuk: yes, just as if you call a missing method on any class.

PR #19, opened 2019-08-20
OPENEUROPA-2172: Implement a custom logger channel factory to instantiate our custom channel class.

imanoleguskiza:
### Description

Currently we are injecting a custom logger channel that only uses the dblog logger. The goal of the PR is to create a new custom LoggerChannelFactory that will return our custom logger channel if prompted, allowing it to use all the loggers available in the site.

### Change log

- Added: PoetryLoggerChannelFactory class.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_translation_poetry/src/Logger/PoetryLoggerChannelFactory.php
+++ b/modules/oe_translation_poetry/src/Logger/PoetryLoggerChannelFactory.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Drupal\oe_translation_poetry\Logger;
+
+use Drupal\Core\Logger\LoggerChannelFactory;
+
+/**
+ * Custom logger channel factory for the Poetry client.
+ *
+ * Since we need a custom logger channel class to process Poetry messages
+ * we also need a custom logger channel factory that will instantiate said
+ * custom logger when prompted.
+ *
+ * @see \Drupal\oe_translation_poetry\Poetry
+ * @see \Drupal\oe_translation_poetry\Logger\PoetryLoggerChannel
+ */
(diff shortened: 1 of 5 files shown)


modules/oe_translation_poetry/src/Logger/PoetryLoggerChannelFactory.php
```diff
@@ -0,0 +1,46 @@
...
+ * @see \Drupal\oe_translation_poetry\Poetry
+ * @see \Drupal\oe_translation_poetry\Logger\PoetryLoggerChannel
+ */
+class PoetryLoggerChannelFactory extends LoggerChannelFactory {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function get($channel) {
+    if ($channel !== 'poetry') {
```
upchuk: This conditional is not needed. We can use multiple channel names for this, no issue. We are the only ones who will use this channel factory.

modules/oe_translation_poetry/src/PoetryFactory.php
```diff
@@ -111,7 +100,7 @@ public function get(string $translator): Poetry {
       $settings = $entity->get('settings');
     }
 
-    return new Poetry($settings, $this->configFactory, $this->loggerChannel, $this->logger, $this->state, $this->entityTypeManager, $this->database, $this->requestStack);
+    return new Poetry($settings, $this->configFactory, $this->loggerFactory, $this->state, $this->entityTypeManager, $this->database, $this->requestStack);
```
upchuk: Lets directly inject into Poetry the channel and not the factory.

PR #23, opened 2019-08-28
OPENEUROPA-2230 - UI for mocking requests.

upchuk:
Aside from the UI for mocks, I also fixed the issue with requesting translations from incorrect source languages.


modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/Controller/PoetryMockController.php
```diff
@@ -108,6 +129,98 @@ public function server(): Response {
...
+      $accepted_languages[] = $language;
+    }
+
+    if ($status === 'CNL') {
+      $cancelled_languages[] = $language;
+    }
+
+    if ($status === 'REF') {
+      $rejected_languages[] = $language;
+    }
```
dxvargas: Better use a "switch" or a "else if".

dxvargas:
There is an error being logged in wathdog each time a notification is mocked, because the request is considered as invalid in `\EC\Poetry\Server::ensureValidRequest`. Reported problems are:
- [0] => Request method should be set to POST.
- [1] => SOAP action header should be defined.

Still the request goes on, so it's not a blocking issue.

PR #25, opened 2019-08-30
OPENEUROPA-2230: Clean translation review for translators.

upchuk:
* Cleaned up a bit the TMGMT job UIs to make it clearer to approve a translation
* Access to the translator role to approve jobs

### DIFF
--- a/runner.yml.dist
+++ b/runner.yml.dist
@@ -47,7 +47,6 @@ commands:
         $settings["poetry.service.password"] = getenv('POETRY_SERVICE_PASSWORD');
         $settings["poetry.notification.username"] = getenv('POETRY_NOTIFICATION_USERNAME');
         $settings["poetry.notification.password"] = getenv('POETRY_NOTIFICATION_PASSWORD');
-        $config['tmgmt.translator.poetry']['settings']['service_wsdl'] = 'http://localhost:8080/build/poetry-mock/wsdl';
     # Generate settings.testing.php, it will be used when running functional tests.
     - { task: "process-php", type: "write", config: "drupal.settings", source: "${drupal.root}/sites/default/default.settings.php", destination: "${drupal.root}/sites/default/settings.testing.php", override: true }
     - task: "append"
--- a/tests/Behat/PoetryTranslationContext.php
+++ b/tests/Behat/PoetryTranslationContext.php
@@ -1,0 +1,271 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_translation\Behat;
+
+use Behat\Behat\Hook\Scope\AfterFeatureScope;
+use Behat\Behat\Hook\Scope\BeforeFeatureScope;
+use Drupal\Core\Database\Query\SelectInterface;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\DrupalExtension\Context\RawDrupalContext;
+use Drupal\node\NodeInterface;
+use Drupal\tmgmt\Entity\Job;
+use EC\Poetry\Messages\Components\Identifier;
+
+/**
+ * Context specific to TMGMT-based poetry translation.
+ */
+class PoetryTranslationContext extends RawDrupalContext {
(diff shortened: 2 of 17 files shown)


tests/features/poetry_translations.feature
```diff
@@ -0,0 +1,62 @@
...
+    Then I should see the link "Translate"
+
+    # Request the translation for two languages
+    When I click "Translate"
+    And I select the languages "Bulgarian, German" in the language list
+    And I press "Request DGT translation for the selected languages"
+    Then I should see "Send request to DG Translation for My title in Bulgarian, German"
+
+    When I fill in "Requested delivery date" with "05/04/2050"
+    And I fill in the the "first" "Author" field with "john"
```
imanoleguskiza: Not to be dealt with here, but I think the Poetry translator should require this fields, since DGT will require them from us as well.

tests/features/poetry_translations.feature
```diff
@@ -0,0 +1,62 @@
...
+    Given the Poetry translation request of "My title" in "Bulgarian, German" gets accepted
+    And I visit "the content administration page"
+    And I click "My title"
+    And I click "Translate"
+    And I should see "Ongoing in Poetry" in the "Bulgarian" row
+    And I should see "Ongoing in Poetry" in the "German" row
+    And I should see "None" in the "Danish" row
+
+    # The translation gets sent from Poetry
+    Given the Poetry translations of "My title" in "Bulgarian, German" get sent by Poetry
```
imanoleguskiza: I think it would be of interest if we only receive the translation of Bulgarian and check that German still has the Ongoing in Poetry label. Otherwise there is little point on sending the german request in the first place.

tests/Behat/PoetryTranslationContext.php
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * This mimics a notification that has come from the Poetry service which
+   * contains the translated content.
+   *
+   * @param string $title
+   *   The node title.
+   * @param string $languages
+   *   The languages.
+   *
+   * @Given the Poetry translation(s) of :title in :languages get sent by Poetry
```
imanoleguskiza: The step wording is a bit weird since on our end there is not sending, I think I would reword it to be something like "he Poetry translation(s) of :title in :languages are received). (Nitpicking here, feel free to ignore.)

runner.yml.dist
```diff
@@ -47,7 +47,6 @@ commands:
         $settings["poetry.service.password"] = getenv('POETRY_SERVICE_PASSWORD');
         $settings["poetry.notification.username"] = getenv('POETRY_NOTIFICATION_USERNAME');
         $settings["poetry.notification.password"] = getenv('POETRY_NOTIFICATION_PASSWORD');
-        $config['tmgmt.translator.poetry']['settings']['service_wsdl'] = 'http://localhost:8080/build/poetry-mock/wsdl';
     # Generate settings.testing.php, it will be used when running functional tests.
```
dxvargas: This is being taken out from here and set only for behat tests.
As a result, when doing manual tests, we have now an error when sending the request.

tests/features/poetry_translations.feature
upchuk: You mean the translator config entity? I tend to disagree a bit. Why should there be site defaults? The form for sending to Poetry has them required, sure, cause that is what is asked by Poetry. But if there are no site defaults, values must be provided at that moment.

tests/features/poetry_translations.feature
upchuk: Brilliant!

runner.yml.dist
upchuk: Yes, we cannot keep that there anymore because it overrides the setting we are putting in Behat. So for manual testing, we need to set that in the settings file ourselves.

tests/features/poetry_translations.feature
imanoleguskiza: Fair enough

PR #31, opened 2019-09-16
OPENEUROPA-2283: Disable Poetry if required configuration is missing.

upchuk:


modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -256,6 +256,50 @@ public function getTranslatorSettings(): array {
...
+   *
+   * If there are missing settings or configuration, Poetry cannot be used
+   * so we use this method to determine this.
+   *
+   * @return bool
+   *   Whether it's available or not.
+   */
+  public function isAvailable(): bool {
+    // Poetry is not available if configuration is missing.
+    $required = [
```
imanoleguskiza: If we are checking that this fields are available then we should make them required in the form as well

upchuk: No, thats not the point actually. We want to disable Poetry if there is no config. This goes in EWCMS, not all sites have multiple languages or even need translations with Poetry. So if these are not configured, there simply won't be Poetry.

PR #42, opened 2019-11-12
Openeuropa 0000 - Poetry README.

upchuk:


modules/oe_translation_poetry/README.md
```diff
@@ -2,14 +2,61 @@
...
 ```
 
-## Requirements
+Then, request Devops to fill those variables on the server with the following values:
 
-If you want to use this submodule, you need to add the following packages to your site:
+* `POETRY_IDENTIFIER_SEQUENCE` - `EWCMS_SEQUENCE`
+* `POETRY_SERVICE_USERNAME` - Username provided by DGT or CEM for accessing the Poetry environment.
+* `POETRY_SERVICE_PASSWORD` - Password provided by DGT or CEM for accessing the Poetry environment.
+* `POETRY_NOTIFICATION_USERNAME` - Username agreed with DGT for authenticating requests made by Poetry to the site (the notifications).
```
dxvargas: For notifications DGT doesn't need to be informed neither to agree with the user nor password.
This information will be sent in the XML of the request and used later by Poetry.

I propose something like we have in NextEuropa:

drupal (user/password) for notifications - in lowercase, limited to 15 characters and different to poetry credential. It must identify uniquely the platform where the translations have to be delivered to. Example: projectname-test/projectname-test or projectname-prod/projectname-prod

modules/oe_translation_poetry/README.md
```diff
@@ -2,14 +2,61 @@
...
+
+### DGT request
+
+To integrate with Poetry, a request has to made to register the site with DGT. See [here](https://github.com/ec-europa/platform-dev/tree/release-2.5/profiles/common/modules/features/nexteuropa_dgt_connector/#requesting-access-to-the-dgt-connector) more information about this. Note: This documentation needs an update.
+
+### Configuration
+
+There are a number of required configuration options needed for the Poetry integration to work. These can be added at this path: `admin/tmgmt/translators/manage/poetry`:
+
+* Poetry service URL - the Poetry endpoint. For the acceptance environment this is `http://intragate.test.ec.europa.eu/DGT/poetry_services/components/poetry.cfc?wsdl`
```
dxvargas: Also production URL should be provided.

upchuk: what is the prod URL?

upchuk: I remove also the test ones. We should not provide these publicly

PR #49, opened 2019-11-29
OPENEUROPA-2535: Poetry HTML export with true #plain_text when needed.

upchuk:


modules/oe_translation_poetry/modules/oe_translation_poetry_html_formatter/tests/fixtures/formatted-content.html
```diff
@@ -20,6 +20,11 @@
           context="[1][translatable_text_field][0][value]"
           -->
           <div class="atom" id="bMV1bdHJhbnNsYXRhYmxlX3RleHRfZmllbGRdWzBdW3ZhbHVl"><h1>This is a heading</h1><p>This is a paragraph</p></div>
+                  <!--
+          label="translatable_text_field_plain"
+          context="[1][translatable_text_field_plain][0][value]"
+          -->
+          <div class="atom" id="bMV1bdHJhbnNsYXRhYmxlX3RleHRfZmllbGRfcGxhaW5dWzBdW3ZhbHVl">plain text field value</div>
```
haringsrob: Test it with actual html inside the field 

'value' => 'plain text field value<br />',

PR #52, opened 2019-12-04
OPENEUROPA-2602: Remove EWCMS references from this module.

dxvargas:


modules/oe_translation_poetry/README.md
```diff
@@ -20,15 +20,15 @@ There are a number of required configuration options needed for the Poetry integ
 
 * Poetry service URL - the Poetry endpoint.
 * Identifier code - Unless requested otherwise, this can be `WEB`
-* Request title prefix - Unless requested otherwise, this can be `EWCMS`
+* Request title prefix - Unless requested otherwise, this can be `OE`
```
upchuk: We should not anymore ship with this but just update the docs to say: ask DGT what title prefix they want.

dxvargas: Fixed.

PR #54, opened 2019-12-04
OPENEUROPA-2252: Allow to request translation updates

dxvargas:
### Description

Allow to request translation updates

### Change log

- Added: Route `oe_translation_poetry.job_queue_checkout_update`
- Added: Form `UpdateTranslationRequestForm`
- Changed: Route `oe_translation_poetry.job_queue_checkout` to `oe_translation_poetry.job_queue_checkout_new`
- Changed: `PoetryNotificationTest::prepareRequestedJobs` to `PoetryTranslationTestBase::createNodeWithRequestedJobs`, doing the same but returning the created node.

### DIFF
--- a/modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
+++ b/modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
@@ -44,7 +44,9 @@ class PoetryRequestTypeEvent extends Event {
    * @param \Drupal\oe_translation_poetry\PoetryRequestType $requestType
    *   The request type constant.
    * @param object|null $jobInfo
-   *   The job information of the current request.
+   *   The job information of the current request. This contains:
+   *     - tjid - the Job ID
+   *     - tjiid -the Job Item ID.
    */
   public function __construct(ContentEntityInterface $entity, PoetryRequestType $requestType, $jobInfo = NULL) {
     $this->entity = $entity;
--- a/modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
+++ b/modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_translation_poetry\Event;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Event class for determining what kind of Poetry request can be made.
+ */
+class PoetryRequestTypeEvent extends Event {
+
+  const EVENT = 'oe_translation_poetry.request_type_event';
+
+  /**
+   * The entity being translated.
+   *
+   * @var \Drupal\Core\Entity\ContentEntityInterface
+   */
+  protected $entity;
+
+  /**
+   * The request type constant.
+   *
+   * @var string
+   */
+  protected $requestType;
+
+  /**
+   * The job information of the current request.
+   *
+   * @var array
+   */
+  protected $jobInfo;
+
+  /**
+   * PoetryRequestTypeEvent constructor.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity being translated.
+   * @param string $requestType
+   *   The request type constant.
+   * @param object|null $jobInfo
+   *   The job information of the current request.
+   */
+  public function __construct(ContentEntityInterface $entity, string $requestType, $jobInfo = NULL) {
+    $this->entity = $entity;
+    $this->requestType = $requestType;
+    $this->jobInfo = $jobInfo;
+  }
+
+  /**
+   * Returns the entity being translated.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface
+   *   The entity.
+   */
+  public function getEntity(): ContentEntityInterface {
+    return $this->entity;
+  }
+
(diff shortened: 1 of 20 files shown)


modules/oe_translation_poetry/src/Poetry.php
```diff
@@ -283,7 +283,7 @@ public function isAvailable(): bool {
    * @return \EC\Poetry\Messages\Components\Identifier
    *   The identifier.
    */
-  protected function getLastIdentifierForContent(ContentEntityInterface $entity): ?Identifier {
+  public function getLastIdentifierForContent(ContentEntityInterface $entity): ?Identifier {
```
dxvargas: This will be used later. Since it makes no harm, I let it be already public.

modules/oe_translation_poetry/src/Plugin/tmgmt/Translator/PoetryTranslator.php
```diff
@@ -335,23 +350,53 @@ public function contentTranslationOverviewAlter(array &$build, RouteMatchInterfa
...
-
     $destination = $entity->toUrl('drupal:content-translation-overview');
     $languages = $this->languageManager->getLanguages();
+
     $unprocessed_languages = $this->getUnprocessedJobsByLanguage($entity);
     $accepted_languages = $this->getAcceptedJobsByLanguage($entity);
     $submitted_languages = $this->getSubmittedJobsByLanguage($entity);
     $translated_languages = $this->getTranslatedJobsByLanguage($entity);
 
+    $request_type = $this->getRequestType($entity, $accepted_languages);
```
dxvargas: Second argument `accepted_languages` needs to be be removed.

modules/oe_translation_poetry/src/Plugin/tmgmt/Translator/PoetryTranslator.php
```diff
@@ -567,6 +618,34 @@ protected function getTranslatedJobsByLanguage(ContentEntityInterface $entity):
...
 
+  /**
+   * Get a list of Poetry jobs that are completed for a given entity.
+   *
+   * These are the jobs which have been translated by Poetry in a certain
+   * request.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity to look jobs for.
+   * @param object $job_info
```
dxvargas: I would prefer to have here either an integer (`$job_info->tjid`) or already the JobInterface $job object.

Like this the $job_info argument is not clear what should be.

modules/oe_translation_poetry/src/Plugin/tmgmt/Translator/PoetryTranslator.php
```diff
@@ -567,6 +618,34 @@ protected function getTranslatedJobsByLanguage(ContentEntityInterface $entity):
     return $result ?? [];
   }
 
+  /**
+   * Get a list of Poetry jobs that are completed for a given entity.
+   *
+   * These are the jobs which have been translated by Poetry in a certain
```
dxvargas: I suggest:

It only returns jobs that were sent in the same request as the given job.

modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
```diff
@@ -0,0 +1,95 @@
...
+  protected $jobInfo;
+
+  /**
+   * PoetryRequestTypeEvent constructor.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity being translated.
+   * @param \Drupal\oe_translation_poetry\PoetryRequestType $requestType
+   *   The request type constant.
+   * @param object|null $jobInfo
```
dxvargas: I would prefer to have here either an integer ($job_info->tjid) or already the JobInterface $job object.

Like this the $job_info argument is not clear what should be.

I put another change request similar to this one after.
But here it's even more important because it's impacting code in other components.

tests/Behat/PoetryTranslationContext.php
```diff
@@ -100,6 +100,25 @@ public function iSelectTheLanguages(string $languages): void {
...
 
+  /**
+   * Asserts that a given language checkbox is checked.
+   *
+   * @param string $language
+   *   The languages.
+   *
+   * @Then the :language language checkbox in the language list is checked.
+   */
+  public function languageCheckboxIsChecked(string $language): void {
```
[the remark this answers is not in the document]
dxvargas: True

tests/features/poetry_translations.feature
```diff
@@ -4,7 +4,7 @@ Feature: Poetry translations
   As a translator
   I need to be able to translate content using the Poetry service
 
-  @cleanup:tmgmt_job @cleanup:tmgmt_job_item @poetry
+  @cleanup:tmgmt_job @cleanup:tmgmt_job_item @poetry @run
```
dxvargas: To be removed.

modules/oe_translation_poetry/src/Event/PoetryRequestTypeEvent.php
upchuk: Mmm, there we need more than just the job object. We need also the Job Item ID..so I will document better the object instead.

tests/Behat/PoetryTranslationContext.php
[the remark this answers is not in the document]
imanoleguskiza: We decided to keep the step for future use

PR #57, opened 2020-01-13
OPENEUROPA-2712: Add validation to checkout form for delivery date.

dxvargas:
### Description

Add validation to checkout form for delivery date, past dates shouldn't be allowed.


modules/oe_translation_poetry/src/Form/PoetryCheckoutFormBase.php
```diff
@@ -190,6 +194,18 @@ public function buildForm(array $form, FormStateInterface $form_state, ContentEn
...
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    // Check delivery date is not in the past.
+    $delivery_date = new \DateTime($form_state->getValue('details')['date']);
+    $today = new \DateTime();
+    if ($delivery_date < $today) {
```
dxvargas: Like this it's not possible to send current day.

upchuk: @dxvargas is that supposed to happen? Send a request for a translation today?

dxvargas: We don't have a requirement to block it. Changing, after agreement.

PR #58, opened 2020-01-13
OPENEUROPA-1766: Fix failing phpunit tests for 2020 problem.

dxvargas:


dxvargas:
> Approve for now, we will have to move that constant to somewhere dynamic as well.

True, I've discussed that in OPENEUROPA-2252, it's under our radar and it should be addressed.

PR #60, opened 2020-01-29
OPENEUROPA-2752: BlockField TMGMT processor.

upchuk:
Submodule that integrates with the Block Field module and makes the block reference field translatable.

### DIFF
--- a/modules/oe_translation_block_field/src/BlockFieldProcessor.php
+++ b/modules/oe_translation_block_field/src/BlockFieldProcessor.php
@@ -9,7 +9,7 @@ use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\tmgmt_content\DefaultFieldProcessor;
 
 /**
- * Field processor for the Block title field.
+ * Field processor for the Block reference field.
  */
 class BlockFieldProcessor extends DefaultFieldProcessor {
 
@@ -29,7 +29,7 @@ class BlockFieldProcessor extends DefaultFieldProcessor {
 
       if (count($field) > 1) {
         // More than one item, add a label for the delta.
-        $data[$delta]['#label'] = t('Delta #@delta', ['@delta' => $delta]);
+        $data[$delta]['#label'] = $this->t('Delta #@delta', ['@delta' => $delta]);
       }
       $data[$delta]['settings__label'] = [
         '#label' => $this->t('Block title'),
@@ -71,7 +71,7 @@ class BlockFieldProcessor extends DefaultFieldProcessor {
   }
 
   /**
-   * Method to set the translated title value back to the block title.
+   * Sets the translated title value onto the block field settings.
    *
    * @param \Drupal\block_field\BlockFieldItemInterface $field_item
    *   The block field item.
--- a/modules/oe_translation_block_field/src/BlockFieldProcessor.php
+++ b/modules/oe_translation_block_field/src/BlockFieldProcessor.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Drupal\oe_translation_block_field;
+
+use Drupal\block_field\BlockFieldItemInterface;
+use Drupal\Core\Field\FieldItemListInterface;
+use Drupal\Core\Render\Element;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\tmgmt_content\DefaultFieldProcessor;
+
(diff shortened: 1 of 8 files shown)


modules/oe_translation_block_field/src/BlockFieldProcessor.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
```
sergepavle: Missed `declare(strict_types = 1);`

modules/oe_translation_block_field/src/BlockFieldProcessor.php
```diff
@@ -0,0 +1,87 @@
...
+class BlockFieldProcessor extends DefaultFieldProcessor {
+
+  use StringTranslationTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function extractTranslatableData(FieldItemListInterface $field) {
+    $data = [];
+    /* @var \Drupal\Core\Field\FieldItemInterface $field_item */
```
sergepavle: Probably should be `\Drupal\Core\Field\FieldDefinitionInterface $field_definition`

modules/oe_translation_block_field/src/BlockFieldProcessor.php
```diff
@@ -0,0 +1,87 @@
...
+      if (count($field) > 1) {
+        // More than one item, add a label for the delta.
+        $data[$delta]['#label'] = $this->t('Delta #@delta', ['@delta' => $delta]);
+      }
+      $data[$delta]['settings__label'] = [
+        '#label' => $this->t('Block title'),
+        '#text' => $property->getValue()['label'],
+        '#translate' => TRUE,
+      ];
+      $data[$delta]['settings__label']['#max_length'] = 255;
```
sergepavle: Probably should be: 

>       $data[$delta]['settings__label'] = [
>         '#label' => $this->t('Block title'),
>         '#text' => $property->getValue()['label'],
>         '#translate' => TRUE,
>        '#max_length' => 255,
>       ];

modules/oe_translation_block_field/src/BlockFieldProcessor.php
```diff
@@ -0,0 +1,87 @@
...
+    /* @var \Drupal\Core\Field\FieldItemInterface $field_item */
+    $field_definition = $field->getFieldDefinition();
+
+    foreach ($field as $delta => $field_item) {
+      $property = $field_item->getProperties()['settings'];
+      $data['#label'] = $field_definition->getLabel();
+
+      if (count($field) > 1) {
+        // More than one item, add a label for the delta.
+        $data[$delta]['#label'] = $this->t('Delta #@delta', ['@delta' => $delta]);
```
sergepavle: is it necessary don't have this with fields where we have only 1 field item?
If simplicity it also makes sense don't use `if (count($field) > 1) {`
optional comment.

modules/oe_translation_block_field/tests/Kernel/BlockFieldTest.php
```diff
@@ -0,0 +1,109 @@
...
+    $values = [
+      'title' => 'My entity',
+      'langcode' => 'en',
+      'user_id' => 1,
+    ];
+    $this->entityTest = EntityTestMul::create($values);
+  }
+
+  /**
+   * Tests tests that the block field is translatable.
```
sergepavle: Looks like typo
```suggestion
   * Tests that the block field is translatable.
```

modules/oe_translation_block_field/tests/Kernel/BlockFieldTest.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
```
sergepavle: Missed `declare(strict_types = 1);`

modules/oe_translation_block_field/tests/Kernel/BlockFieldTest.php
```diff
@@ -0,0 +1,109 @@
...
+      'langcode' => 'en',
+      'user_id' => 1,
+    ];
+    $this->entityTest = EntityTestMul::create($values);
+  }
+
+  /**
+   * Tests tests that the block field is translatable.
+   */
+  public function testBlockFieldTranslation() {
```
sergepavle:
```suggestion
  public function testBlockFieldTranslation(): void {
```

PR #63, opened 2020-01-31
OPENEUROPA-2251: Allow to request languages addition.

dxvargas:
### Description

Allow to add languages to a request ongoing in Poetry.


modules/oe_translation_poetry/src/Form/AddLanguagesRequestForm.php
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $node
+   *   The node entity to use when generating the title.
+   *
+   * @return \Drupal\Core\StringTranslation\TranslatableMarkup
+   *   The title for the form page.
+   */
+  public function getPageTitle(ContentEntityInterface $node = NULL): TranslatableMarkup {
+    $queue = $this->queueFactory->get($node);
+    $entity = $queue->getEntity();
```
imanoleguskiza: We already have the node entity in the method parameters so there is no need to get it again.

modules/oe_translation_poetry/oe_translation_poetry.module
```diff
@@ -91,30 +91,52 @@ function oe_translation_poetry_form_tmgmt_content_translate_form_alter(&$form, F
...
+    }
+  }
+
+  // Add a button to request new languages. This can only happen while there
+  // is an accepted ongoing request.
+  if (!$accepted_languages) {
+    return;
   }
+
+  $form['actions']['request_new_languages'] = [
```
imanoleguskiza: We probably also want extra validation for this action. Currently if I click the "Add extra languages" button without selecting any new language, I end up in an "Access denied" page.

modules/oe_translation_poetry/oe_translation_poetry.module
```diff
@@ -91,30 +91,52 @@ function oe_translation_poetry_form_tmgmt_content_translate_form_alter(&$form, F
...
-
   $languages = array_merge($accepted_languages, $translated_languages, $completed_languages);
-  foreach (array_keys($languages) as $langcode) {
-    $form['languages'][$langcode] = [
-      '#type' => 'checkbox',
-      '#disabled' => TRUE,
-      '#value' => 1,
-    ];
+
+  // If we are talking about an update request, we force the submission in
```
imanoleguskiza: This comment is confusing, I think it needs rewording.

modules/oe_translation_poetry/src/Form/AddLanguagesRequestForm.php
```diff
@@ -0,0 +1,98 @@
...
+
+namespace Drupal\oe_translation_poetry\Form;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\Core\Url;
+
+/**
+ * Form for requesting new translations.
```
imanoleguskiza: This is not very representative. How about something like "Form for requesting new languages for an existing translation request"

modules/oe_translation_poetry/src/Plugin/tmgmt/Translator/PoetryTranslator.php
```diff
@@ -376,24 +389,33 @@ public function contentTranslationOverviewAlter(array &$build, RouteMatchInterfa
...
+    // the request ID from one of the accepted jobs in order to filter the
+    // completed jobs and include only the ones in this request. For the
+    // translated jobs we don't have to filter by request ID because those
+    // still contain the `poetry_state` value which indicates they are part
+    // of the last request for that content.
+    if ($accepted_languages) {
       $accepted_job_info = reset($accepted_languages);
       /** @var \Drupal\tmgmt\JobInterface $job */
       $job = $this->entityTypeManager->getStorage('tmgmt_job')->load($accepted_job_info->tjid);
       $completed_languages = $this->getCompletedJobsByLanguage($entity, $job);
```
imanoleguskiza: I know it was not introduced in this PR but this method shouldn't get the $job as a parameter, it should get the request_id.

modules/oe_translation_poetry/src/Form/AddLanguagesRequestForm.php
upchuk: Not exactly. The queue provides us with the exact revision of the entity. And its title is the one we want to use.

PR #64, opened 2020-02-07
OPENEUROPA-2796: Add cancelled information in translation page

dxvargas:
### Description

Add cancelled languages information in translation page. It's shown both for request and language cancellation.


tests/Behat/PoetryTranslationContext.php
```diff
@@ -178,6 +180,78 @@ public function thePoetryTranslationRequestGetAccepted(string $title, string $la
...
+      $language = current($this->getLanguagesFromNames($node_hash['language']));
+      $status = $node_hash['status'];
+
+      $language_send = [
+        'code' => strtoupper($language->getId()),
+        'date' => '30/08/2050 23:59',
+        'accepted_date' => '30/09/2050 23:59',
+      ];
+
+      if ($request_status == 'Ongoing') {
```
imanoleguskiza: The $request_status_send vairable shouldn't be set inside the for loop.

tests/Behat/PoetryTranslationContext.php
```diff
@@ -178,6 +180,78 @@ public function thePoetryTranslationRequestGetAccepted(string $title, string $la
...
+        $request_status_send = 'ONG';
```
imanoleguskiza: I think a bettter name would be $demand_status, since $request_status_send doesn't say much about that we are storing.

tests/features/poetry_translations.feature
```diff
@@ -157,3 +157,90 @@ Feature: Poetry translations
...
+    And I fill in "Secretary" with "john"
+    And I fill in "Contact" with "john"
+    And I fill in the the "first" "Responsible" field with "john"
+    And I fill in the the "second" "Responsible" field with "john"
+    And I fill in the the "second" "Author" field with "john"
+    And I fill in "Requester" with "john"
+    And I press "Send request"
+    Then I should see "The request has been sent to DGT."
+    And I should not see "Cancelled in Poetry" in the "Bulgarian" row
+    And I should not see "Cancelled in Poetry" in the "Czech" row
```
imanoleguskiza: For completeness sake we should probably also add:
And I should see "Submitted to Poetry" in the "Spanish" row

tests/Behat/PoetryTranslationContext.php
dxvargas: I've choose "demand" word here before, then @upchuk have changed it. And I think it's ok like this.
We use request for this, demand seems more a french wording like we see in other places of the service. We're talking about the request status so it's an appropriate terminology.

PR #66, opened 2020-02-11
OPENEUROPA-2795: Test we can receive translation updates while a job is active

imanoleguskiza:
### Description

We need to be able to receive and update translations while a job is active. 
Add tests and the possibility to manually test this functionality.
### Change log

- Added: Tests for updating translations and new option to mock it.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/Controller/PoetryMockController.php
+++ b/modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/Controller/PoetryMockController.php
@@ -203,8 +203,10 @@ class PoetryMockController extends ControllerBase {
     $item = reset($items);
     $data = \Drupal::service('tmgmt.data')->filterTranslatable($item->getData());
     foreach ($data as $field => &$info) {
+      // Check whether this is a new translation or not by checking for a
+      // stored translation for the field.
       if (isset($info['#translation'])) {
-        $info['#text'] = $info['#translation']['#text'] . ' UPDATE';
+        $info['#text'] = $info['#translation']['#text'] . ' OVERRIDDEN';
       }
       else {
         $info['#text'] .= ' - ' . $tmgmt_job->getTargetLangcode();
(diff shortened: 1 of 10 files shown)


modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/EventSubscriber/ContentTranslationOverviewAlterSubscriber.php
```diff
@@ -135,6 +136,16 @@ public function alterOverview(ContentTranslationOverviewAlterEvent $event): void
           ),
         ];
       }
+      if (isset($translated_jobs[$language->getId()])) {
+        $links['update_translation'] = [
```
dxvargas: I would avoid calling this action update. We already have a "translation update" that has nothing to do with this operation. I suggest translation override.

modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/Controller/PoetryMockController.php
```diff
@@ -203,7 +203,13 @@ public function sendTranslationNotification(JobInterface $tmgmt_job): Response {
     $item = reset($items);
     $data = \Drupal::service('tmgmt.data')->filterTranslatable($item->getData());
     foreach ($data as $field => &$info) {
-      $info['#text'] .= ' - ' . $tmgmt_job->getTargetLangcode();
+      if (isset($info['#translation'])) {
+        $info['#text'] = $info['#translation']['#text'] . ' UPDATE';
```
dxvargas: We already have a "translation update" that has nothing to do with this operation.
I suggest here ' OVERRIDDEN'.

modules/oe_translation_poetry/modules/oe_translation_poetry_mock/src/EventSubscriber/ContentTranslationOverviewAlterSubscriber.php
imanoleguskiza: I'm ok with calling the link "translation_override" but I think the title needs to be more descriptive about what we are doing. Something like "Receive translation update (mock)" would do it I think?

dxvargas: So we have already
* Accept job (mock)
* Translate job (mock)
* Cancel job (mock)
What about having now one of
* Translate job (mock)
* Translate job again (mock)
* Retranslate job (mock)

PR #73, opened 2020-02-27
OPENEUROPA-2608: Generate new breadcrumb only if entity exists.

22Alexandra:

### DIFF
--- a/src/Plugin/tmgmt/Translator/PermissionTranslator.php
+++ b/src/Plugin/tmgmt/Translator/PermissionTranslator.php
@@ -537,12 +537,7 @@ class PermissionTranslator extends TranslatorPluginBase implements ApplicableTra
     array_pop($links);
     array_pop($links);
 
-    $item_type_storage = $this->entityTypeManager->getStorage($item_type);
-    if (empty($item_type_storage)) {
-      return;
-    }
-
-    $entity = $item_type_storage->load($job_item->getItemId());
+    $entity = $this->entityTypeManager->getStorage($item_type)->load($job_item->getItemId());
     $links[] = $entity->toLink(NULL, 'drupal:content-translation-overview');
     $breadcrumb = new Breadcrumb();
     $breadcrumb->setLinks($links);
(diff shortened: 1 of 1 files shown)


src/Plugin/tmgmt/Translator/PermissionTranslator.php
```diff
@@ -535,7 +536,13 @@ public function localTaskItemBreadcrumbAlter(Breadcrumb &$breadcrumb, RouteMatch
     array_pop($links);
     array_pop($links);
     array_pop($links);
-    $entity = $this->entityTypeManager->getStorage('node')->load($job_item->getItemId());
+
+    $item_type_storage = $this->entityTypeManager->getStorage($item_type);
+    if (empty($item_type_storage)) {
+      return;
```
upchuk:
```
  /**
   * Creates a new storage instance.
   *
   * @param string $entity_type_id
   *   The entity type ID for this storage.
   *
   * @return \Drupal\Core\Entity\EntityStorageInterface
   *   A storage instance.
   *
   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
   *   Thrown if the entity type doesn't exist.
   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
   *   Thrown if the storage handler couldn't be loaded.
   */
  public function getStorage($entity_type_id);
```

src/Plugin/tmgmt/Translator/PermissionTranslator.php
```diff
@@ -525,6 +525,7 @@ public function localTaskItemBreadcrumbAlter(Breadcrumb &$breadcrumb, RouteMatch
     /** @var \Drupal\tmgmt_local\LocalTaskItemInterface $local_task_item */
     $local_task_item = $route_match->getParameter('tmgmt_local_task_item');
     $job_item = $local_task_item->getJobItem();
+    $item_type = $job_item->getItemType();
```
imanoleguskiza: We could do this inside the try or just before it so in case the method returns early we don't set it for nothing.

PR #81, opened 2020-03-20
OPENEUROPA-2943: Upgrade to PHP 7.3.

sergepavle:
### Description

[Insert description here]

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,6 +68,7 @@
         }
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "discard-changes": true
     }
 }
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -68,6 +69,7 @@
         }
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "discard-changes": true
```
ademarco: Why is this necessary?

sergepavle: Exist issue on drone side (on composer install).
For not completely reasons we have on lower dependencies patched already drupal/coder module and we have error like that:

> RuntimeException]                                                                  
> Source directory /test/oe_translation/vendor/drupal/coder has uncommitted changes.  

By these settings, we enforce composer discard changes and discard changes (patched files)

PR #97, opened 2020-05-07
OPENEUROPA-2907: Add EntityRevisionWithTypeItem field type.

22Alexandra:


src/Plugin/Validation/Constraint/EntityRevisionWithTypeConstraint.php
```diff
@@ -0,0 +1,33 @@
...
+namespace Drupal\oe_translation\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * EntityRevisionWithType constraint.
+ *
+ * @Constraint(
+ *   id = "EntityRevisionWithType",
+ *   label = @Translation("Entity Revision WithT ype", context = "Validation")
```
imanoleguskiza, line 14: Small typo here.

src/Plugin/Field/EntityRevisionWithTypeItemListInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_translation\Plugin\Field;
+
+/**
+ * Defines field item list classes for EntityRevisionWithTypeItem field items.
+ */
+interface EntityRevisionWithTypeItemListInterface {
```
imanoleguskiza, line 10: This probably needs to extend FieldItemListInterface.

src/Plugin/Field/FieldFormatter/EntityRevisionWithTypeFormatter.php
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * Constructs a EntityRevisionWithTypeFormatter object.
```
imanoleguskiza, line 37: Constructs an EntityRevisionWithTypeFormatter object.

src/Plugin/Field/FieldType/EntityRevisionWithTypeItem.php
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function getConstraints() {
+    $constraint_manager = \Drupal::typedDataManager()->getValidationConstraintManager();
+    $constraints = parent::getConstraints();
+
+    $constraints[] = $constraint_manager->create('ComplexData', [
+      'entity_type' => [
+        'EntityRevisionWithType' => [],
```
imanoleguskiza, line 70: Why do we not have constraints that validate that the id or revision id actually exist? As it is I could store whatever I want as long as its a string.

tests/Kernel/EntityRevisionWithTypeFieldTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_translation\Kernel;
+
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+
+/**
+ * Tests the EntityRevisionWithType field type, widget and formatter.
```
imanoleguskiza, line 11: We are not really testing the widget though, we would need to use the form for that.

src/Plugin/Field/FieldType/EntityRevisionWithTypeItem.php
upchuk: It's true. But it won't fail. It will reference a not found entity which is fine. Entity references can do that we well. We need just a simple field for storage not go crazy in functionality now.

PR #99, opened 2020-05-08
OPENEUROPA-2906: Add Translation Synchronisation field type.

22Alexandra:


src/Plugin/Field/FieldWidget/TranslationSynchronisationWidget.php
```diff
@@ -0,0 +1,118 @@
...
+      '#options' => static::getSyncTypeOptions(),
+      '#default_value' => isset($items[$delta]->type) ? $items[$delta]->type : NULL,
+    ];
+
+    // Get the selector to build the paths for type field.
+    $parents = $element['#field_parents'];
+    $parents[] = $this->fieldDefinition->getName();
+    $parents[] = $element['#delta'];
+    $selector = array_shift($parents);
+    if ($parents) {
```
imanoleguskiza, line 46: I don't understand this very well: previously we assigned two new elements to the $parents array so even if we shift the first one the second will always be there so $parents will never be empty.

tests/FunctionalJavascript/TranslationTestBase.php
```diff
@@ -0,0 +1,66 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_translation\FunctionalJavascript;
+
+use Drupal\FunctionalJavascriptTests\WebDriverTestBase;
+
+/**
+ * Base class for functional javsacript tests.
+ */
+class TranslationTestBase extends WebDriverTestBase {
```
imanoleguskiza, line 12: We don't use this base class anywhere. Also it should probably be abstract (if we keep it)

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
```diff
@@ -0,0 +1,164 @@
...
+    $this->getSession()->getPage()->pressButton('Save');
+    /** @var \Drupal\node\NodeInterface $node */
+    $node = $this->entityTypeManager->getStorage('node')->load(1);
+    $this->assertTrue($node->get('translation_sync')->isEmpty());
+    $this->assertEmpty($node->get('translation_sync')->getValue());
+
+    // Edit the node again and configure it to be Manual.
+    $this->drupalGet($node->toUrl('edit-form'));
+    $this->getSession()->getPage()->selectFieldOption('Select type', 'manual');
+    $this->getSession()->wait(10);
```
imanoleguskiza, line 112: Not a huge fan of adding this wait here. I understand you want to make sure the configuration is not loaded but I wonder if it is even necessary?

tests/Kernel/TranslationSynchronisationFieldTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Tests the field type and formatter.
+   */
+  public function testTranslationSynchronisationItemField(): void {
+    /** @var \Drupal\node\NodeStorageInterface $node_storage */
+    $node_storage = $this->container->get('entity_type.manager')->getStorage('node');
+    $date = new \DateTime('2020-05-08');
+    $tests = [
+      'manual' => [
```
imanoleguskiza, line 50: We are missing an empty case.

tests/Kernel/TranslationSynchronisationFieldTest.php
```diff
@@ -0,0 +1,96 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_translation\Kernel;
+
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\Entity\FieldStorageConfig;
+use Drupal\oe_translation\Plugin\Field\FieldFormatter\TranslationSynchronisationFormatter;
+
+/**
+ * Tests the Translation Synchronisation field type, widget and formatter.
```
imanoleguskiza, line 12: We don't test the widget here.

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
```diff
@@ -0,0 +1,164 @@
...
+
+    // Assert the select is correct.
+    $select = $assert->selectExists('translation_sync[0][type]');
+    $assert->optionExists('Select type', 'None');
+    $assert->optionExists('Select type', 'Manual');
+    $assert->optionExists('Select type', 'Automatic with minimum threshold');
+    $this->assertEquals('None', $select->find('xpath', "//option[@selected='selected']")->getText());
+
+    // Assert we don't see a configuration fieldset.
+    $this->assertFalse($this->getSession()->getPage()->find('css', '#edit-translation-sync-0-configuration')->isVisible());
```
imanoleguskiza, line 100: I would avoid using the ids to find the configuration. Why not use the selectExists method like before?

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
```diff
@@ -0,0 +1,164 @@
...
+    $this->assertEquals($expected_values, $node->get('translation_sync')->first()->getValue());
+
+    // Edit the node again and configure it to be Automatic without a date.
+    $this->drupalGet($node->toUrl('edit-form'));
+    $this->getSession()->getPage()->selectFieldOption('Select type', 'automatic');
+    $assert->waitForField('Languages');
+    $this->assertTrue($this->getSession()->getPage()->find('css', '#edit-translation-sync-0-configuration')->isVisible());
+    $this->assertTrue($this->getSession()->getPage()->find('css', '#edit-translation-sync-0-configuration-languages')->isVisible());
+    $this->assertTrue($this->getSession()->getPage()->find('css', '#edit-translation-sync-0-configuration-date')->isVisible());
+    $this->getSession()->getPage()->selectFieldOption('Languages', 'en', TRUE);
```
imanoleguskiza, line 130: We could also check that the languages field is required but the date field is not.

tests/FunctionalJavascript/TranslationTestBase.php
upchuk: We will use it soon so let's not delete it. Are the other base test classes abstract? If so, yes, make them abstract

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
upchuk: No, the wait here is to give enough time for the #states to actually hide or show. We wanna make sure that we dont have a false positive and in fact the states does not show anything. Where the states actually shows something, we do a wait for element instead of a random amount of time.

tests/Kernel/TranslationSynchronisationFieldTest.php
upchuk: True

tests/Kernel/TranslationSynchronisationFieldTest.php
upchuk: No, we test it in the functional one!

upchuk: Indeed, comment is outdated.

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
upchuk: Ah yes, true

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
upchuk: It's fine, we assert the entire container cause thats what the states affects.

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
imanoleguskiza: If we wanted to be super precise we could add a condition to the wait so it ends if it finds the container, but probably not worh the effort.

tests/FunctionalJavascript/TranslationSynchronisationFieldWidgetTest.php
upchuk: It's ok, we cover this in the kernel

tests/FunctionalJavascript/TranslationTestBase.php
imanoleguskiza: The other base tests are not abstract neither, even though I think they should be. Can be fixed at a later stage.

PR #102, opened 2020-05-18
OPENEUROPA-2909: Personal & Organisational Contacts field types.

22Alexandra:


modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
```diff
@@ -0,0 +1,191 @@
...
+    parent::setUp();
+
+    $this->installSchema('node', ['node_access']);
+
+    $this->container->get('entity_type.manager')->getStorage('node_type')->create([
+      'type' => 'page',
+      'name' => 'Page',
+    ])->save();
+
+    FieldStorageConfig::create([
```
imanoleguskiza, line 37: We could create a FieldManagementTrait to create fields and avoid duplicating code here and on PoetryRequestIdFieldTest.

imanoleguskiza: Even better we can check if similar traits already exist.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
```diff
@@ -0,0 +1,191 @@
...
+    ])->save();
+  }
+
+  /**
+   * Tests the Organisational field type and formatter.
+   */
+  public function testOrganisationalContactItemField(): void {
+    /** @var \Drupal\node\NodeStorageInterface $node_storage */
+    $node_storage = $this->container->get('entity_type.manager')->getStorage('node');
+    $tests = [
```
imanoleguskiza, line 68: This seems like it should be handled by a data provider method.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
```diff
@@ -0,0 +1,191 @@
...
+      'type' => 'oe_translation_poetry_organisation_contact',
+    ])->save();
+
+    FieldConfig::create([
+      'field_name' => 'organisational_contact_info',
+      'entity_type' => 'node',
+      'bundle' => 'page',
+    ])->save();
+
+    FieldStorageConfig::create([
```
imanoleguskiza, line 49: If each of the methods tests one of the fields, should we create both as part of the setup or rather create each field in each method?

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
```diff
@@ -0,0 +1,191 @@
...
+      $build = $builder->viewField($node->get('organisational_contact_info'));
+      $output = $this->container->get('renderer')->renderRoot($build);
+      $this->assertContains(implode('<br />', $values), (string) $output);
+    }
+  }
+
+  /**
+   * Tests the Personal field type and formatter.
+   */
+  public function testPersonalContactItemField(): void {
```
imanoleguskiza, line 124: Both test methods do exactly the same thing, I believe we should be able to merge both an use a data provider method that provides the field name, the values and the expected empty() result.

modules/oe_translation_poetry/tests/Functional/ContactFieldsWidgetTest.php
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+
+    $this->nodeStorage = \Drupal::entityTypeManager()->getStorage('node');
+
+    foreach ($this->getTestCases() as $case) {
+      FieldStorageConfig::create([
```
imanoleguskiza, line 45: I think its better if we are consistent and either use the static ::create methods or call \Drupal::entityTypeManager()->getStorage manually, but not both.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
imanoleguskiza, line 37: No need to call the static method since we can get the entity_type.manager service from the container.

modules/oe_translation_poetry/tests/Functional/ContactFieldsWidgetTest.php
```diff
@@ -0,0 +1,193 @@
...
+    $this->drupalLogin($user);
+  }
+
+  /**
+   * Prepare the test cases for the two field types.
+   *
+   * @return array
+   *   The test cases.
+   */
+  protected function getTestCases(): array {
```
imanoleguskiza, line 83: This looks and behaves exactly like a data provider method, so why not make it a data provider?

modules/oe_translation_poetry/src/Plugin/Field/FieldFormatter/OrganisationalContactFormatter.php
```diff
@@ -0,0 +1,41 @@
...
+ *   label = @Translation("Organisational Contact Formatter"),
+ *   field_types = {"oe_translation_poetry_organisation_contact"}
+ * )
+ */
+class OrganisationalContactFormatter extends FormatterBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewElements(FieldItemListInterface $items, $langcode) {
```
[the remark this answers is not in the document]
upchuk: No, here we dont

modules/oe_translation_poetry/src/Plugin/Field/FieldType/OrganisationalContactItem.php
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Defines the 'Organisational Contact' field type.
+ *
+ * @FieldType(
+ *   id = "oe_translation_poetry_organisation_contact",
+ *   label = @Translation("Organisational Contact"),
+ *   category = @Translation("OpenEuropa"),
+ *   default_widget = "oe_translation_poetry_organisation_contact_widget",
+ *   default_formatter = "oe_translation_poetry_organisation_contact_formatter"
```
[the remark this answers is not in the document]
upchuk: I was thinking about this as well, but we might have to ourselves. But I agree, let's put no UI and then if we have to we can enable it temporarily. This is not a field type meant for the UI. Also, @22Alexandra please do for for the sync field type and entity rev with item one as well.

modules/oe_translation_poetry/src/Plugin/Field/FieldWidget/ContactWidgetBase.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Base class for contact field type widgets.
+ */
+abstract class ContactWidgetBase extends WidgetBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
+    $types = $items->first()->contactTypes();
```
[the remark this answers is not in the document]
upchuk: It doesn't matter, but yeah, it's cleaner.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: Yes, I wanted to ask @22Alexandra to do this, then I took over and forgot :))

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: No, we don't wanna run new tests for each of the case. No data provider in non-Unit tests.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: Ah yes indeed, let's do that.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: Yes, we can merge the tests, but no data provider.

modules/oe_translation_poetry/tests/Functional/ContactFieldsWidgetTest.php
upchuk: Yes, apart from setting the node storage as class property, let's use static everywhere.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: Yes, but it doesn't matter so much. Plus we refactor to use a trait

modules/oe_translation_poetry/tests/Functional/ContactFieldsWidgetTest.php
upchuk: becasue we dont want a test run for each of the cases

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: In the end, I don't think there is much we can save..it's simple field creation.

modules/oe_translation_poetry/tests/Kernel/ContactFieldsTest.php
upchuk: We merge the tests so irrelevant

PR #105, opened 2020-05-25
OPENEUROPA-2908: Create Translation Request entity type.

22Alexandra:


oe_translation.permissions.yml
```diff
@@ -0,0 +1,17 @@
+administer translation requests:
+  title: 'Administer translation requests'
+  restrict access: true
+administer translation request types:
+  title: 'Administer translation request types'
+  restrict access: true
+access translation request overview:
```
imanoleguskiza, line 7: This permission and the next one seem to serve no purpose. As a user with this permission I would assume I can access the translation request collection route, but I can't

22Alexandra: They are needed just that I was using the the core route provider. I've added the route provider for the entity so now the permission are working.

PR #118, opened 2020-11-13
EWPP-516-2: Update docker images.

22Alexandra:


sergepavle (approved):

nagyad (approved):

composer.json
```diff
@@ -14,7 +14,7 @@
         "behat/mink-selenium2-driver": "1.4.x-dev as 1.3.x-dev",
         "composer/installers": "~1.5",
         "drupal-composer/drupal-scaffold": "~2.2",
-        "drupal/address": "^1.8",
+        "drupal/address": "^1.9",
```
brummbar, line 17: Same as the other PR, I think it's enough to fix the selector without updating the dependency.

brummbar (changes requested):

.drone.yml
```diff
@@ -14,12 +14,13 @@ services:
...
   mysql:
-    image: percona/percona-server:5.6
+    image: registry.fpfis.eu/fpfis/sql:percona-5.7
+    command: --innodb-log-file-size=1G --max_allowed_packet=1G --innodb-buffer-pool-size=512M --wait_timeout=3000 --net_write_timeout=3000 --log_error_verbosity=3
     environment:
       - MYSQL_ALLOW_EMPTY_PASSWORD=yes
   selenium:
     # Newer Selenium Chrome images are not supported at the moment by behat/mink-selenium2-driver.
-    image: selenium/standalone-chrome-debug:3.141.59-oxygen
+    image: registry.fpfis.eu/fpfis/selenium:standalone-chrome-debug-3.141.59-oxygen
```
brummbar, line 23: Also here let's use the non-debug image.

brummbar (changes requested):

composer.json
22Alexandra: The selector would have the ["."](https://github.com/openeuropa/oe_translation/pull/118/files#diff-1a9ec38cbc32142ff1173d6e1fb95478c601d02f31d195a6eae628eaa0f5fb92R160) on 1.8 and doesn't have it on 1.9. I chose to fix the selector without the dot and to update the lowest version to 1.9 so the selector is always the same.

brummbar: Ah makes sense! It's also a dev requirement only... We can go like this, perfect.

sergepavle (approved):

sergepavle (approved):

nagyad (approved):

brummbar (approved):

PR #128, opened 2021-02-11
EWPP-769: Allowing a retry on accepting a translation.

upchuk:


modules/oe_translation_poetry/oe_translation_poetry.module
```diff
@@ -183,9 +183,17 @@ function oe_translation_poetry_tmgmt_job_item_access(EntityInterface $entity, $o
   }
 
   if ($job->get('poetry_state')->value !== PoetryTranslator::POETRY_STATUS_TRANSLATED) {
-    // Forbid access to job item updates if they have not been translated by
-    // Poetry.
-    return AccessResult::forbidden();
+    // Forbid access to job item updates if they are not waiting to be accepted
+    // or if there is no translation on the target entity from this job item.
```
imanoleguskiza, line 187: This is a bit misleading I think, what is happening is more like "Forbid access to job item updates if they are not waiting to be accepted or if the source entity for this item is no longer available.".

imanoleguskiza (changes requested):

22Alexandra (approved):

imanoleguskiza (approved):

PR #136, opened 2021-04-08
EWPP-959: Fix on the storing of Poetry error messages in the job.

upchuk:


upchuk:
I could not provide a failing test commit because when the SQL crashes in the background, the test execution simply stops and the test stays green :)

Fefaine:
This won't work either ?
https://github.com/openeuropa/oe_translation/blob/20f165b46b9beeb93bae49de4e3636b92a4efa21/modules/oe_translation_poetry/tests/Functional/PoetryMockTest.php#L92
I was checking with Tamas on Covid19 and we get this error 
SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'message' at row 1: INSERT INTO @tmgmt_message....

upchuk:
@Fefaine What? 

That is what this PR solves. Those errors.

22Alexandra (approved):

imanoleguskiza (approved):

PR #142, opened 2021-05-10
EWPP-1103: Move test metatag field into separate test module.

nagyad:


tests/Functional/TranslationTestBase.php
```diff
@@ -36,6 +36,7 @@ class TranslationTestBase extends BrowserTestBase {
     'options',
     'oe_translation',
     'oe_translation_test',
+    'oe_translation_metatag_test',
     'metatag',
```
sergepavle, line 40: As I understand 'metatag' should be removed from the list of modules in the test.

tests/modules/oe_translation_test/oe_translation_test.install
```diff
@@ -21,7 +21,6 @@ function oe_translation_test_uninstall() {
     'node.ott_top_level_paragraphs',
     'node.ott_inner_paragraph_field',
     'node.ott_top_level_paragraph_field',
-    'node.field_metatag'
```
sergepavle, line 24: Should we move deleting of the field in oe_translation_metatag_test module?

tests/Functional/TranslationTestBase.php
```diff
@@ -36,6 +36,7 @@ class TranslationTestBase extends BrowserTestBase {
     'options',
     'oe_translation',
     'oe_translation_test',
+    'oe_translation_metatag_test',
```
sergepavle, line 39: Do we really need to enable oe_translation_metatag_test for all Functional tests?
Maybe we will do enabling in tests where we really do testing?

tests/Functional/TranslationTestBase.php
nagyad: No not from the test.

tests/Functional/TranslationTestBase.php
nagyad: It was, why would we change now?

sergepavle: It is an optional comment. As far as we did reorganising of dependencies I just throught that it would be reasonable also to simplify dependencies in tests.

sergepavle (approved):

sergepavle (dismissed):

upchuk (changes requested):

upchuk (approved):

PR #160, opened 2022-04-12
EWPP-2104: Handling concurrent unprocessed poetry jobs.

upchuk:


modules/oe_translation_poetry/src/Plugin/tmgmt/Translator/PoetryTranslator.php
```diff
@@ -648,6 +650,19 @@ public function contentTranslationOverviewAlter(array &$build, RouteMatchInterfa
...
   public function submitPoetryTranslationRequest(array &$form, FormStateInterface $form_state): void {
+    /** @var \Drupal\Core\Entity\ContentEntityInterface $entity */
+    $entity = $form_state->get('entity');
+    $entity = $entity->isDefaultTranslation() ? $entity : $entity->getUntranslated();
+    // Check if there are any unprocessed jobs for this entity and prevent
+    // concurrent requests from duplicating jobs. We check if we haven't yet
+    // already disabled the jobs because when we do that, the button turns into
+    // one meant to finalise the same jobs.
+    $unprocessed_jobs = $this->getUnprocessedJobsByLanguage($entity);
+    if (!empty($unprocessed_jobs) && !isset($form['#unprocessed_jobs_disabled'])) {
```
imanoleguskiza, line 661: Here we have a problem because we never enter this condition, if a second concurrent user tries to submit a request for the same language, $unprocessed_jobs will not be empty but $form['#unprocessed_jobs_disabled'] will be set, so the execution continues as normal

imanoleguskiza (changes requested):

imanoleguskiza (dismissed):

22Alexandra (dismissed):

22Alexandra (approved):

yenyasinn (approved):

PR #164, opened 2022-05-19
EWPP-2254: Update TranslationRequest entity type.

22Alexandra:


src/Entity/TranslationRequest.php
```diff
@@ -160,101 +161,29 @@ public function setTargetLanguageCodes(array $target_language_codes): Translatio
...
+  public function getData(): string {
+    return Json::decode($this->get('data')->value);
   }
 
   /**
    * {@inheritdoc}
    */
-  public function setRequestStatus(string $request_status): TranslationRequestInterface {
-    return $this->set('request_status', $request_status);
+  public function setData(string $data): TranslationRequestInterface {
```
upchuk, line 185: Well, if you encode into a json string, then we need to pass an array here no? Not a string...

This shows also that this test is useless:

```
    $data_for_translation = $this->randomString();
    $translation_request->setData($data_for_translation);
    $this->assertEquals($data_for_translation, $translation_request->getData());
```

upchuk (changes requested):

upchuk (approved):

PR #165, opened 2022-05-19
EWPP-2251: Add translator providers service.

22Alexandra:


src/TranslatorProviders.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_translation;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+
+/**
+ * Checks if an entity type is exposed to any translator provider.
```
upchuk, line 10: Service used for determining if an entity type is exposed to our translation system.

src/TranslatorProviders.php
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Retrieves the entity type oe_translation_translators configuration.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type.
+   *
+   * @return array|bool
+   *   Array containing the configuration, FALSE if there is none.
+   */
+  protected function getTranslators(EntityTypeInterface $entity_type): ?array {
```
upchuk, line 56: Let's always return arrays, empty if none.

upchuk (changes requested):

upchuk (approved):

PR #166, opened 2022-05-20
EWPP-2266: Translation dashboard skeleton + old code removal.

upchuk:


src/EventSubscriber/TranslationOverviewRequestSubscriber.php
```diff
@@ -83,13 +94,12 @@ public function onRequest(GetResponseEvent $event): void {
     }
 
     $entity_type = $matches[1];
-    if (!$this->entityTypeManager->getDefinition($entity_type)) {
+    if (!$this->entityTypeManager->hasDefinition($entity_type)) {
```
22Alexandra, line 97: Shall we replace deprecated class `GetResponseEvent` with `RequestEvent` (the `$event` parameter type)?

src/Plugin/Derivative/TranslationLocalTasks.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
```
22Alexandra, line 2:
```suggestion

declare(strict_types = 1);

```

src/Plugin/Derivative/TranslationLocalTasks.php
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * The translation providers service.
+   *
+   * @var \Drupal\oe_translation\TranslatorProvidersInterface
+   */
+  protected $translatorProviders;
+
+  /**
+   * Creates a FieldUiLocalTask object.
```
22Alexandra, line 33:
```suggestion
   * Creates a TranslationLocalTasks object.
```

src/Plugin/Derivative/TranslationLocalTasks.php
```diff
@@ -0,0 +1,82 @@
...
+    return new static(
+      $base_plugin_id,
+      $container->get('oe_translation.translator_providers'),
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
```
22Alexandra, line 58:
```suggestion
  public function getDerivativeDefinitions($base_plugin_definition): array {
```

src/TranslatorProvidersInterface.php
```diff
@@ -44,4 +44,23 @@ public function hasRemote(EntityTypeInterface $entity_type): bool;
...
   public function getRemotePlugins(EntityTypeInterface $entity_type): array;
 
+  /**
+   * Determines whether the entity type has any translator plugins.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type.
+   *
+   * @return bool
+   *   TRUE if it has at least a remote translator, FALSE otherwise.
```
22Alexandra, line 54:
```suggestion
   *   TRUE if it has any translator, FALSE otherwise.
```

src/Plugin/Derivative/TranslationLocalTasks.php
```diff
@@ -0,0 +1,82 @@
...
+      $this->derivatives["$entity_type_id.dashboard"] = [
+        'route_name' => "entity.$entity_type_id.content_translation_overview",
+        'title' => $this->t('Dashboard'),
+        'parent_id' => "content_translation.local_tasks:entity.$entity_type_id.content_translation_overview",
+        'weight' => 1,
+      ] + $base_plugin_definition;
+
+      // @todo change this to the proper route for local translations.
+      $this->derivatives["$entity_type_id.local_translations"] = [
+        'route_name' => "entity.$entity_type_id.canonical",
```
22Alexandra, line 72: Looks a bit weird to have it in the canonical, I would move it to `entity.$entity_type_id.content_translation_overview` to have it on the same page with the Dashboard but I guess it's fine here too until we have the route..

22Alexandra (changes requested):

src/Plugin/Derivative/TranslationLocalTasks.php
upchuk: Parent doesn't have it, not our interface.

src/Plugin/Derivative/TranslationLocalTasks.php
upchuk: Yeah, it's fine like this.

src/EventSubscriber/TranslationOverviewRequestSubscriber.php
upchuk: Yes, thanks! I also included a test for this redirect which I had deleted.

22Alexandra (approved):

hernani (approved):

PR #170, opened 2022-06-01
EWPP-2269: Add Remote translation provider plugin type.

22Alexandra:


modules/oe_translation_remote/tests/src/Functional/RemoteTranslationTest.php
```diff
@@ -5,7 +5,7 @@
 namespace Drupal\Tests\oe_translation_remote\Functional;
 
 use Drupal\node\Entity\Node;
-use Drupal\Tests\oe_translation\Functional\TranslationTestBase;
+use Drupal\Tests\oe_translation\FunctionalJavascript\TranslationTestBase;
```
upchuk, line 8: You made this a functional JS test but left it in the Functional namespace?

modules/oe_translation_remote/tests/src/Functional/RemoteTranslationTest.php
```diff
@@ -36,4 +37,66 @@ public function testRemoteTranslationRoute(): void {
...
+    $this->assertSession()->elementAttributeContains('css', 'select#edit-plugin', 'required', 'required');
+
+    // Assert different configuration form based on selected plugin.
+    $this->getSession()->getPage()->selectFieldOption('Plugin', 'Remote one');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->assertSession()->pageTextContains('Plugin configuration for Remote one');
+    $this->assertSession()->pageTextContains('This plugin does not have any configuration options.');
+    $this->getSession()->getPage()->selectFieldOption('Plugin', 'Remote two');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->assertSession()->pageTextContains('Plugin configuration for Remote two');
```
upchuk, line 76: Assert the page no longer contains the config form for remote one.

modules/oe_translation_remote/tests/src/Functional/RemoteTranslationTest.php
```diff
@@ -36,4 +37,66 @@ public function testRemoteTranslationRoute(): void {
...
+    $this->getSession()->getPage()->pressButton('Save');
+    $this->assertSession()->pageTextContains('Created the Test provider Remote Translator Provider.');
+
+    // Edit the provider.
+    $this->clickLink('Edit');
+    $this->assertSession()->fieldValueEquals('Name', 'Test provider');
+    $this->assertSession()->fieldValueEquals('A test configuration string', 'Plugin configuration.');
+    $this->getSession()->getPage()->fillField('Name', 'Test provider edited');
+    $this->getSession()->getPage()->pressButton('Save');
+    $this->assertSession()->pageTextContains('Saved the Test provider edited Remote Translator Provider.');
```
upchuk, line 93: Do also a load of the config entity and assert all the exported values are in the correct format.

tests/src/Functional/TranslationTestBase.php
```diff
@@ -65,7 +65,6 @@ protected function setUp(): void {
     $role = $this->entityTypeManager->getStorage('user_role')->load('oe_translator');
     $permissions = $role->getPermissions();
     $permissions[] = 'administer menu';
-    $permissions[] = 'edit any page content';
```
upchuk, line 68: Why did you remove this?

tests/src/FunctionalJavascript/TranslationTestBase.php
```diff
@@ -53,10 +58,16 @@ protected function setUp() {
...
+    $this->drupalPlaceBlock('page_title_block', ['region' => 'content']);
+    $this->drupalPlaceBlock('local_tasks_block', ['region' => 'content']);
+    $this->drupalPlaceBlock('local_actions_block', ['region' => 'content']);
+
     /** @var \Drupal\user\RoleInterface $role */
     $role = $this->entityTypeManager->getStorage('user_role')->load('oe_translator');
     $permissions = $role->getPermissions();
     $permissions[] = 'administer menu';
+    $permissions[] = 'edit any page content';
+    $permissions[] = 'administer site configuration';
```
upchuk, line 70: Well, we cannot add this permission that essentially allows them to do anything. It's no longer a translator then. You need to keep this user a regular translator, which may create some content, fine, but not do everything. Make another user where you need the proper role for creating provider entities.

modules/oe_translation_remote/tests/src/Functional/RemoteTranslationTest.php
```diff
@@ -36,4 +37,66 @@ public function testRemoteTranslationRoute(): void {
     $this->assertSession()->pageTextContains('The remote translations overview page.');
   }
 
+  /**
+   * Test the remote translation provider configuration.
+   */
+  public function testRemoteTranslationProvider(): void {
```
upchuk, line 43: Name the tests a bit better, this is very vague and the test doesn't do this. The test tests that creation of a translation provider config. So something like `testRemoteTranslationProviderForm` is better.

tests/src/Functional/TranslationTestBase.php
22Alexandra: Because I previously added it as it was needed for `testRemoteTranslationRoute` but now that I switched to functional JS, I moved the permission in that base class. The permission itself is only temporary needed anyway as it's supposed to be replaced later.

upchuk (dismissed):

hernani:
image
Only problem found relates to dashboard being visible in view tab . Also translate/view tab seem both active.

upchuk:
@hernani Like I said, this is now fixed with the merge of the local translations.

upchuk (approved):

imanoleguskiza (approved):

PR #172, opened 2022-06-12
EWPP-2309: Update path for new entity types and new modules.

22Alexandra:


upchuk (changes requested):
The definitions need to be hardcoded in the update path and not loaded dynamically. They need to match what we are installing at this exact moment in time.

upchuk (approved):

imanoleguskiza (approved):

PR #186, opened 2022-11-21
EWPP-2803: Language with status and local translation bundle class.

upchuk:


config/schema/oe_translation.schema.yml
```diff
@@ -30,3 +30,14 @@ oe_translation.oe_translation_request_type.*:
     label:
       type: label
       label: 'Label'
+
+field.value.oe_translation_language_with_status:
```
22Alexandra, line 34: We need to update the schema for existing sites cause the Local translations dashboard is not accessible now.

22Alexandra (changes requested):
Tests for the new field type?

config/schema/oe_translation.schema.yml
upchuk: It's not that. It was caused from before, fixed by fixing the entity definition creation in the update hook.

22Alexandra (dismissed):

22Alexandra (approved):

hernani (approved):

PR #207, opened 2023-03-08
EWPP-3146: Removing TMGMT.

upchuk:
The scope of this PR is to remove TMGMT and move the code we need to oe translation.


src/TranslationSourceFieldProcessor/TranslationSourceFieldProcessorInterface.php
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Interface for translation source field processors.
+ *
+ * @ingroup oe_translation
+ */
+interface TranslationSourceFieldProcessorInterface {
+
+  /**
+   * Extracts the translatatable data structure from the given field.
```
22Alexandra, line 17:
```suggestion
   * Extracts the translatable data structure from the given field.
```

22Alexandra (approved):

PR #220, opened 2023-03-20
EWPP-2856: Preview remote translation requests.

22Alexandra:


modules/oe_translation_remote/src/Controller/RemoteContentTranslationPreviewController.php
```diff
@@ -0,0 +1,88 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function previewRequest(TranslationRequestInterface $oe_translation_request, string $language): array {
+    // Generate the translation preview for the request in the current language.
+    $language = $this->languageManager->getCurrentLanguage()->getId();
+    $translation = $this->getTranslationPreview($oe_translation_request, $language, $oe_translation_request->bundle());
```
upchuk, line 62: You are passing the `$oe_translation_request` entity, you don't need a separate argument for the bundle, you can just get it from the entity/

modules/oe_translation_remote/src/Form/RemoteTranslationReviewForm.php
```diff
@@ -192,6 +199,25 @@ public function synchronise(array &$form, FormStateInterface $form_state): void
...
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function preview(array &$form, FormStateInterface $form_state) {
+    /** @var \Drupal\oe_translation_local\TranslationRequestLocal $translation_request */
+    $translation_request = $this->entity;
+
+    $language = $form_state->get('language');
+    $url = $translation_request->toUrl('preview');
+    $url->setRouteParameter('language', $language->id());
```
upchuk, line 216: Is this needed? You are setting the URL option the language and in the controller you are checking the current language, not the route parameter.

src/Controller/ContentTranslationPreviewController.php
```diff
@@ -132,23 +116,41 @@ public function previewRequestTitle(TranslationRequestInterface $oe_translation_
...
+  protected function getTranslationPreview(TranslationRequestInterface $oe_translation_request, string $language, string $bundle = 'local'): ContentEntityInterface {
+    // Get the entity being translated.
+    $entity = $oe_translation_request->getContentEntity();
+    if (!$entity) {
+      throw new \Exception('There is no entity currently being translated.');
+    }
+
+    // Retrieve the translation data.
+    if ($bundle !== 'local') {
+      $data = $oe_translation_request->getTranslatedData();
```
upchuk, line 138: What is this? `getTranslatedData ` is literally from another module. The entire point of this ticket is to move the logic of getting the data out of the entity where it belongs. By default it's `$oe_translation_request->getData();`, otherwise, the `$oe_translation_request->getTranslatedData();` belongs in the remote module

This block here defeats the entire purpose of the PR. You could have left everything exactly the same and just add this. No need for any other controller anything..

upchuk (changes requested):

modules/oe_translation_remote/src/Form/RemoteTranslationReviewForm.php
22Alexandra: Yes, the language is a required parameter for the route. I removed the option and the check for the current language in controller.

src/Controller/ContentTranslationPreviewController.php
```diff
@@ -107,26 +98,40 @@ public function previewRequest(TranslationRequestInterface $oe_translation_reque
...
+   *
+   * @param \Drupal\oe_translation\Entity\TranslationRequestInterface $oe_translation_request
+   *   The translation request.
+   * @param string $language
+   *   The language to preview in.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface
+   *   Translated entity.
+   */
+  protected function extractPreviewData(TranslationRequestInterface $oe_translation_request, string $language): ContentEntityInterface {
```
upchuk: You are not returning data but an entity. So `extractPreviewTranslation`

hernani (dismissed):

upchuk (dismissed):

hernani (approved):

upchuk (approved):

PR #229, opened 2023-03-24
EWPP-3194: Various UX improvements.

upchuk:


modules/oe_translation_epoetry/src/Plugin/views/field/TargetLanguagesWithTooltip.php
```diff
@@ -0,0 +1,192 @@
...
+      $count += count($grouped[TranslationRequestEpoetryInterface::STATUS_LANGUAGE_REVIEW]);
+    }
+
+    // Include the synced ones.
+    if (isset($grouped[TranslationRequestEpoetryInterface::STATUS_LANGUAGE_SYNCHRONISED])) {
+      $count += count($grouped[TranslationRequestEpoetryInterface::STATUS_LANGUAGE_SYNCHRONISED]);
+    }
+
+    // Include also the languages which have been accepted locally, but only
+    // if the there is translation data.
```
22Alexandra:
```suggestion
    // if there is translation data.
```

modules/oe_translation_epoetry/src/Plugin/views/field/TargetLanguagesWithTooltip.php
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @param \Drupal\oe_translation_epoetry\TranslationRequestEpoetryInterface $request
+   *   The request.
+   * @param string $language
+   *   The language.
+   *
+   * @return bool
+   *   Whether a given translation has arrived from ePoetry.
+   */
+  protected function isLanguageArrived(TranslationRequestEpoetryInterface $request, string $language): bool {
```
22Alexandra:
```suggestion
  protected function hasLanguageArrived(TranslationRequestEpoetryInterface $request, string $language): bool {
```

modules/oe_translation_epoetry/src/Plugin/views/filter/TargetLanguagesFilter.php
```diff
@@ -0,0 +1,61 @@
...
+   * The language manager.
+   *
+   * @var \Drupal\Core\Language\LanguageManagerInterface
+   */
+  protected $languageManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, Connection $connection, LanguageManagerInterface $languageManager) {
```
22Alexandra: What do we need the database for? Doesn't seem to be used.

22Alexandra (commented):
The view link of the ongoing remote translations leads to the request's canonical page and we cannot go back to the dashboard from there. I think we should use the `entity.$entity_type_id.remote_translation` route.
view link

upchuk:
@22Alexandra 

> The view link of the ongoing remote translations leads to the request's canonical page and we cannot go back to the dashboard from there. I think we should use the entity.$entity_type_id.remote_translation route.

There can be multiple requests on the remote_translation route so we cannot link there always. However, if you enable breakdcrumbs in your theme, you can use those to navigate (check in ewcms)

hernani (approved):

22Alexandra (approved):

PR #233, opened 2023-04-03
EWPP-3205: Various UX improvements.

upchuk:


modules/oe_translation_corporate_workflow/tests/src/Functional/CorporateWorkflowTranslationTest.php
```diff
@@ -409,10 +409,35 @@ public function testDefaultLocalModeratedTranslation(): void {
...
+      '2.0.0 / validated',
+      'Edit draft translationDelete',
+    ];
+    $this->assertLocalOngoingRequests($expected);
+
     // Publish the node before finalizing the translation.
     $this->moderateNode($node, 'published');
     $revision_ids = $node_storage->revisionIds($node);
     $this->assertCount(10, $revision_ids);
+    // Assert the dashboard contains to ongoing requests.
```
22Alexandra:
```suggestion
    // Assert the dashboard contains the ongoing request.
```

modules/oe_translation_epoetry/oe_translation_epoetry.module
```diff
@@ -317,3 +319,31 @@ function oe_translation_epoetry_preprocess_table__remote_language_list(array &$v
...
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ */
+function oe_translation_epoetry_oe_translation_request_presave(TranslationRequestInterface $request) {
+  if ($request->bundle() !== 'epoetry') {
+    return;
+  }
+  // Mark the request as finalized if the ePoetry status is Executed AND all
+  // languages are Cancelled. This is because Cancelled languages cannot be
+  // continued anymore so the request needs be finished.
```
22Alexandra, line 332:
```suggestion
  // continued anymore so the request needs to be finished.
```

modules/oe_translation_epoetry/oe_translation_epoetry.module
```diff
@@ -317,3 +319,31 @@ function oe_translation_epoetry_preprocess_table__remote_language_list(array &$v
...
+  // We do this here because the ePoetry notifications
+  // can come in any order so we have to check every time the request is saved.
+  $epoetry_status = $request->getEpoetryRequestStatus();
+  if ($epoetry_status !== TranslationRequestEpoetryInterface::STATUS_REQUEST_EXECUTED) {
+    return;
+  }
+
+  $languages = $request->getTargetLanguages();
+  foreach ($languages as $language_with_status) {
+    if ($language_with_status->getStatus() !== TranslationRequestEpoetryInterface::STATUS_LANGUAGE_CANCELLED) {
```
22Alexandra: We should mark the request as finished if all the requested languages are either cancelled, accepted or synchronised (basically sent from ePoetry), not just cancelled.

22Alexandra (changes requested):

modules/oe_translation_epoetry/oe_translation_epoetry.module
upchuk: No, if a language is accepted locally, it means the request should stil be open, as the user still needs to sync it

upchuk: Fixed to take into account the synced languages

22Alexandra (approved):

hernani (approved):

PR #234, opened 2023-04-06
EWPP-3205: Adding Rejected product status.

upchuk:


modules/oe_translation_epoetry/oe_translation_epoetry.module
```diff
@@ -327,26 +328,64 @@ function oe_translation_epoetry_oe_translation_request_presave(TranslationReques
...
-  // Mark the request as finalized if the ePoetry status is Executed AND all
-  // languages are Cancelled. This is because Cancelled languages cannot be
-  // continued anymore so the request needs be finished.
-  // We do this here because the ePoetry notifications
+
+  $languages = $request->getTargetLanguages();
+
+  // Mark the request as Translated if all languages are either Cancelled,
+  // Rejected or in Review. This is because Cancelled and Rejected languages
+  // cannot be continued anymore so the they should allow the request to be
```
22Alexandra: Same below on line 357.
```suggestion
  // cannot be continued anymore so they should allow the request to be
```

modules/oe_translation_epoetry/src/RequestFactory.php
```diff
@@ -302,13 +302,18 @@ protected function toRequestDetails(TranslationRequestEpoetryInterface $request)
...
-      ->setComment($entity->toUrl('revision')->setAbsolute()->toString())
       ->setLinguisticSections($linguistic_sections);
     $request_details->setOriginalDocument($original_document);
 
+    $comment = '';
     if ($request->getMessage()) {
-      $request_details->setComment($request->getMessage());
+      $comment .= trim($request->getMessage(), '.');
     }
+    if ($comment !== "") {
```
22Alexandra, line 312: Use single quotes.

22Alexandra (changes requested):
I found this issue (might be related to the mock only):

I have an ongoing request where the 1st language is Rejected, the 2nd one is Cancelled and after the 3rd language is Translated the request status is updated to Translated but the `newTranslationRequestForm` is called. Shouldn't it be called only after the request is marked as Finished?
new request
 
Then, I was able to add new languages (Hungarian) to this request:
new language to Translated request

While I already have two ongoing requests I can start another one:
new request while 2 are already ongoing

upchuk:
@22Alexandra RE the comment with the screenshot, yes, it's a bit weird, but there are 2 things:

1. When the request is Translated on our side, we need to still allow to add new languages. As per DGT request
2. Yes, a bit weird that if all languages arrive (or some of them get rejected/cancelled), we can the make an entirely new request as well (before updating the content). But that has been like that frm the beginning and it's ok. The API allows us to do that and can unblock certain use cases in which there is a mistake (for example, they rejected or cancelled a translation by mistake). Typically, translators wont do this because they won't have the need to do.

modules/oe_translation_corporate_workflow/tests/src/Functional/CorporateWorkflowTranslationTest.php
```diff
@@ -862,6 +862,80 @@ public function testModerationStateSync(): void {
     $assert_translation_state($moderation_state, 'validated');
   }
 
+  /**
+   * Tests that synchronising translations doesn't change the default revision.
+   *
+   * The tested use case is having a translation started from a Published
+   * version that becomes unpublished, followed by syncing that ongoing
+   * translation. We ensure that when we sync onto that Published version which
```
hernani, line 870: Published version, which

modules/oe_translation_corporate_workflow/tests/src/Functional/CorporateWorkflowTranslationTest.php
```diff
@@ -862,6 +862,80 @@ public function testModerationStateSync(): void {
     $assert_translation_state($moderation_state, 'validated');
   }
 
+  /**
+   * Tests that synchronising translations doesn't change the default revision.
+   *
+   * The tested use case is having a translation started from a Published
```
hernani, line 868: The test starts with a translation started from a Published version that becomes unpublished, followed by syncing that ongoing translation.

22Alexandra (approved):

hernani (approved):

PR #236, opened 2023-04-17
Fix queries when using table prefix

lisotton:
The queries implemented in `oe_translation_poetry_post_update_0002` are failing because the tables are hard-coded. 
For websites that use table prefix it will fail because the table will not exist. 

The table names should be wrapped with `{}`, so Drupal will prepend the prefix before running the query.


22Alexandra (approved):

upchuk:
hey @lisotton thanks, moved the commit to https://github.com/openeuropa/oe_translation/pull/237 cause I had to rebase.

PR #238, opened 2023-04-18
Change AvailableLanguagesAlterEvent to extend Symfony event

lisotton:


upchuk:
it's the phpsoap client reques the problem, not this.

PR #241, opened 2023-04-19
EWPP-3034: Show warning message upon submitting a request with weekend deadline.

22Alexandra:


modules/oe_translation_epoetry/tests/src/FunctionalJavascript/EpoetryTranslationTest.php
```diff
@@ -2547,6 +2547,51 @@ public function testAutomatedRequestStatusUpdate(): void {
...
+   * Tests that warning message is shown for requests with a weekend deadline.
+   */
+  public function testWeekendDeadlineWarning(): void {
+    $node = $this->createBasicTestNode();
+    $this->drupalGet($node->toUrl('drupal:content-translation-overview'));
+    $this->clickLink('Remote translations');
+    // Select a language.
+    $this->getSession()->getPage()->checkField('Bulgarian');
+    // Select a weekend deadline.
+    $this->getSession()->getPage()->fillField('translator_configuration[epoetry][deadline][0][value][date]', '08/26/2023');
```
upchuk: @22Alexandra This is not a good date to use. In a few months, this test will fail because the date will be in the past

upchuk (changes requested):

modules/oe_translation_epoetry/tests/src/FunctionalJavascript/EpoetryTranslationTest.php
```diff
@@ -2555,6 +2555,51 @@ public function testAutomatedRequestStatusUpdate(): void {
     }
   }
 
+  /**
+   * Tests that warning message is shown for requests with a weekend deadline.
+   */
+  public function testWeekendDeadlineWarning(): void {
```
upchuk: I'm sorry for all this extra work but for this tiny assertion you can just add to the `testEpoetrySingleTranslationFlow` where we also assert things like the mandatory language and the max length on the message. So, you can just add a few lines there instead of having an entire test running just for this.

upchuk (changes requested):

sergepavle (dismissed):

upchuk (approved):

sergepavle (approved):

Issue #245, opened 2023-04-26
oe_translation_post_update_0002() fails if oe_translator role was deleted on website


donquixote:
If on a website the oe_translator role no longer exists, the post update hook will fail.

```php
function oe_translation_post_update_0002(&$sandbox = NULL) {
  $role = Role::load('oe_translator');
  $role->grantPermission('accept translation request');
  $role->grantPermission('sync translation request');
  $role->save();
}
```

> Error: Call to a member function grantPermission() on null in oe_translation_post_update_0002()

donquixote:
An alternative could be to assign these new permissions to any role that currently has another permission.
But I don't know if this can be safely assumed to be the right thing for all websites. And I don't know which prior permission we would use for this.

upchuk:
@donquixote Thanks. Will be fixed in alpha2. We cannot assign to a random role. So we will just bail out if the role is missing.

donquixote:
Would be nice to have something we can use as a patch.

upchuk:
@donquixote It's fixed here: https://github.com/openeuropa/oe_translation/pull/247
You can use this branch as patch

donquixote:
> @donquixote It's fixed here: #247 You can use this branch as patch

There is another one that needs fixing for `oe_translation_local_install()`.

upchuk:
Indeed. Fixed as well

PR #247, opened 2023-04-27
EWPP-3205: Allow to update incoming remote translations.

upchuk:


modules/oe_translation_local/src/Form/LocalTranslationRequestForm.php
```diff
@@ -310,48 +309,6 @@ public function preview(array &$form, FormStateInterface $form_state) {
...
-   * @param array $values
-   *   Nested array of values to set.
-   * @param bool $replace
-   *   (optional) When TRUE, replaces the structure at the provided key instead
-   *   of writing into it.
-   *
-   * @return array
-   *   The updated data.
-   */
-  protected function updateData($key, array $data, array $values, bool $replace = FALSE) {
```
upchuk, line 335: This has been moved to the trait, no changes made to it.

src/Form/TranslationRequestForm.php
```diff
@@ -46,7 +46,10 @@ public function save(array $form, FormStateInterface $form_state) {
    * @see TranslationFormTrait::translationFormElement()
    */
   public static function validateMaxLength(array $element, FormStateInterface &$form_state): void {
-    if (isset($element['#max_length']) && ($element['#max_length'] < mb_strlen($element['#value']))) {
+    // The value can be nested in a rare case in which the element is a
+    // text_format.
+    $value = $element['#value'] ?? $element['value']['#value'];
```
upchuk, line 51: In case we have a text format element with max lenght, the value is nested, so we need to account for this. No test added now because discovered another (bigger issue) which we might fix after, so the text would come then.

22Alexandra (dismissed):

22Alexandra:
We should fix this issue https://github.com/openeuropa/oe_translation/issues/245 within the PR.

22Alexandra (approved):

hernani (approved):

PR #254, opened 2023-05-16
EWPP-2965: Drupal 10 compatibility.

brummbar:


brummbar:
Support for Drupal 10 will be added only for 2.x releases, as [the Poetry client](https://github.com/ec-europa/oe-poetry-client) is being deprecated and it would need a major overhaul to make it work with Drupal 10.

PR #260, opened 2023-05-31
EWPP-3327: Adding translation column groups to the typed link field t…

upchuk:


22Alexandra (dismissed):
We should probably remove the core patch, because it only applies when requiring composer to install the lowest versions, and add an entry in README.md for anyone that still uses 9.4.x to apply it in their project.

hernani (dismissed):

brummbar (approved):

22Alexandra (approved):

PR #263, opened 2023-06-20
EWPP-3388: Ensuring that ePoetry server errors are logged correctly in the request log.

upchuk:


modules/oe_translation_epoetry/modules/oe_translation_epoetry_mock/src/Controller/MockController.php
```diff
@@ -107,6 +107,7 @@ public function server(Request $request): Response {
       $wrapper = file_get_contents($this->moduleExtensionList->getPath('oe_translation_epoetry_mock') . '/fixtures/error_wrapper.xml');
       $wrapper = str_replace(['@code', '@string'], $error_response, $wrapper);
       $response = new Response($wrapper);
+      $response->setStatusCode(500);
```
upchuk, line 110: This was the key problem before when testing. It used to work because the mock would not return a 500, but a 200 with an error response. Now it replicates correctly what the real life scenario does. The tests, as they are, confirm that we get the message from the error. Since the exception message contains that truncated value.

hernani (approved):

22Alexandra (approved):

PR #280, opened 2023-11-13
Use same source for dossier part when using the Mock

lisotton:
Maybe you have another motivation to use a separated state to store the parts for mock, but this is causing issues when using copy of production database with mock to test the integration with Epoetry.

It happens that it uses by default the state `oe_translation_epoetry_mock.last_part_for_number`, which will start from `1`, but use the same dossier number, so when the mock is used to update the status, it will not update the right ePoetry request in Drupal, instead it will update an old request, which the part is 1. 

The workaround to make this work, is to go in Structure > Remote Translator Provider list > Edit ePoetry and check "Reset current dossier". 

I'm wondering, why not use the same source as the module "oe_translation_epoetry" to return the part?


lisotton:
I have noticed that the `createNewVersion` is also using an state to retrieve the next version. The problem is that if you use a copy of the database from PROD, it will start from the version `1`, which might happen to already have that version. So we will end up having to requests with the same Request ID. The if you try to use the mock to update the request, it will send the notifications to another request.

PR #298, opened 2024-03-18
D8AGE-287: Create TranslationRequest bundle for CDT.

intelektron:
### Description

The new TranslationRequest bundle with initial configuration and bundle class.

### Change log

- Added: bundle and test


modules/oe_translation_cdt/src/TranslationRequestCdt.php
```diff
@@ -0,0 +1,184 @@
...
+   */
+  public function setCdtId(string $value): TranslationRequestCdtInterface {
+    $this->set('cdt_id', $value);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setRequestStatusFromCdt(string $cdt_status): TranslationRequestCdtInterface {
```
intelektron, line 35: I did not introduce a separate field, or a mapper here.
I considered this approach much simplier.

modules/oe_translation_cdt/src/TranslationRequestCdt.php
```diff
@@ -0,0 +1,184 @@
...
+    };
+    $this->setRequestStatus($target_status);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function updateTargetLanguageStatusFromCdt(string $langcode, string $cdt_status): void {
+    $target_status = match($cdt_status) {
```
intelektron, line 49: The mapping is primitive, but IMO it covers our needs.

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.contact_usernames.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_translation
+id: oe_translation_request.contact_usernames
+field_name: contact_usernames
```
drishu, line 7: dont think the epoetry and cdt will ever need to be installed at the same time, but at least lets name it better.. like contact_names

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.phone_number.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_translation
+id: oe_translation_request.phone_number
+field_name: phone_number
+entity_type: oe_translation_request
+type: string
```
drishu, line 9: 👍  The ticket says number, but I also think that's wrong.
Also if any extra phone validation/constraints are needed I think we should have a dedicated ticket for that

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.correlation_id.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_translation
+id: oe_translation_request.correlation_id
+field_name: correlation_id
+entity_type: oe_translation_request
+type: string
```
drishu, line 9: From the swagger API I see "2019/000365" as example correlation id, so again I think string is the way to go :thumbsup:

modules/oe_translation_cdt/config/install/field.field.oe_translation_request.cdt.request_status.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.oe_translation_request.request_status
```
drishu, line 5: 👍

modules/oe_translation_cdt/config/install/field.field.oe_translation_request.cdt.target_languages.yml
```diff
@@ -0,0 +1,20 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.oe_translation_request.target_languages
```
drishu, line 5: 👍

modules/oe_translation_cdt/config/install/field.field.oe_translation_request.cdt.request_status.yml
```diff
@@ -0,0 +1,20 @@
...
+    - oe_translation.oe_translation_request_type.cdt
+  module:
+    - options
+id: oe_translation_request.cdt.request_status
+field_name: request_status
```
drishu, line 10: ticket asked for status but I dont mind, it gives consistency

modules/oe_translation_cdt/src/TranslationRequestCdt.php
```diff
@@ -0,0 +1,184 @@
...
+
+use Drupal\oe_translation\Entity\TranslationRequest;
+use Drupal\oe_translation_remote\RemoteTranslationRequestEntityTrait;
+use Drupal\oe_translation_remote\TranslationRequestRemoteInterface;
+
+/**
+ * A CDT bundle class for oe_translation_request entities.
+ */
+final class TranslationRequestCdt extends TranslationRequest implements TranslationRequestCdtInterface {
+  use RemoteTranslationRequestEntityTrait;
```
drishu, line 15: if we use this then I think we also need the translator_provider

modules/oe_translation_cdt/src/TranslationRequestCdt.php
```diff
@@ -0,0 +1,184 @@
...
+    $this->set('cdt_id', $value);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setRequestStatusFromCdt(string $cdt_status): TranslationRequestCdtInterface {
+    $target_status = match($cdt_status) {
+      'COMP' => TranslationRequestRemoteInterface::STATUS_REQUEST_TRANSLATED,
```
drishu, line 37: these constants, COMP, CANC, CMP, should be defined in the cdt client and imported here
From the docs its a bit dubious, the callback definition can send CMP or CNC, but we can get back COMP

modules/oe_translation_cdt/src/TranslationRequestCdtInterface.php
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * An interface for CDT bundle class for oe_translation_request entities.
+ */
+interface TranslationRequestCdtInterface extends TranslationRequestRemoteInterface {
+
+  /**
+   * Returns the CDT ID.
+   *
+   * @return string|null
+   *   The CDT ID including request year, or NULL if not set.
```
drishu: lets mention the format or an example rather than "including request year , or NULL if not set."

modules/oe_translation_cdt/src/TranslationRequestCdtInterface.php
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Set the request status based on CDT code.
+   *
+   * @param string $cdt_status
+   *   The CDT status code.
+   */
+  public function setRequestStatusFromCdt(string $cdt_status): TranslationRequestCdtInterface;
+
+  /**
+   * Set the language status based on CDT code.
```
drishu: Updates the..

modules/oe_translation_cdt/src/TranslationRequestCdtInterface.php
```diff
@@ -0,0 +1,201 @@
...
+   *
+   * @param string $langcode
+   *   The Drupal language code.
+   * @param string $cdt_status
+   *   The CDT status code.
+   */
+  public function updateTargetLanguageStatusFromCdt(string $langcode, string $cdt_status): void;
+
+  /**
+   * Returns the comments.
```
drishu: Gets..

modules/oe_translation_cdt/src/TranslationRequestCdtInterface.php
```diff
@@ -0,0 +1,201 @@
...
+   * @param string $value
+   *   The confidentiality.
+   *
+   * @return TranslationRequestCdtInterface
+   *   The current request.
+   */
+  public function setConfidentiality(string $value): TranslationRequestCdtInterface;
+
+  /**
+   * Returns the contacts.
```
drishu: Same everywhere, use the verb from the function name

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestBundleClassTest.php
```diff
@@ -0,0 +1,92 @@
...
+    assert($request instanceof TranslationRequestCdtInterface);
+
+    $request->setCdtId('test_cdt_id');
+    $this->assertEquals('test_cdt_id', $request->getCdtId());
+
+    $request->setRequestStatusFromCdt('COMP');
+    $this->assertEquals(TranslationRequestRemoteInterface::STATUS_REQUEST_TRANSLATED, $request->getRequestStatus());
+
+    $request->updateTargetLanguageStatusFromCdt('en', 'CMP');
+    $this->assertEquals(TranslationRequestRemoteInterface::STATUS_LANGUAGE_REVIEW, $request->getTargetLanguage('en')?->getStatus());
```
drishu, line 53:
```suggestion
    $this->assertEquals(TranslationRequestRemoteInterface::STATUS_LANGUAGE_REVIEW, $request->getTargetLanguage('en')->getStatus());
```
Is this a typo?

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/oe_translation_cdt_mock.info.yml
```diff
@@ -0,0 +1,8 @@
+name: OpenEuropa Translation CDT Mock
```
drishu: I dont see why we should add this now

drishu (changes requested):
Please improve the PR title, make it start with a verb

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.contact_usernames.yml
intelektron: I named it after CDT's property name `contact_user_names` (reference: https://github.com/openeuropa/cdt-client/blob/72e9de7752ebeda0872526c4176681dfcc524c6d/src/Model/Request/Translation.php#L21). After the Wednesday call we agreed that the CDT and ePoetry won't be installed together, but it would be good to avoid naming conflicts anyway.

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.phone_number.yml
intelektron: CDT does not validate it, even `abc` will work.

modules/oe_translation_cdt/config/install/field.storage.oe_translation_request.correlation_id.yml
intelektron: The ID that you mentioned is actually permanent ID, the sample correlation ID is `1xWrUG`, so it's also the string.

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestBundleClassTest.php
intelektron: I used the null-safe operator, because the exception thrown in case if `getTargetLanguage('en')` is `NULL` was misleading.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/oe_translation_cdt_mock.info.yml
intelektron: I must have accidentally select it, will remove.

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestBundleClassTest.php
intelektron: And the other thing is that the PHPStan pipeline will fail without `?`, because of the type mismatch

```
------ ------------------------------------------------ 
  Line   TranslationRequestBundleClassTest.php           
 ------ ------------------------------------------------ 
  53     Cannot call method getStatus() on               
         Drupal\oe_translation\LanguageWithStatus|null.  
 ------ ------------------------------------------------ 
```

modules/oe_translation_cdt/src/TranslationRequestCdt.php
intelektron: I planned to add this connection in D8AGE-290, as the translation provider is not available in the context of D8AGE-287.

modules/oe_translation_cdt/src/TranslationRequestCdt.php
intelektron: These values are defined in the callback schema, we talked that they shouldn't be hardcoded anywhere in our code. Perhaps we had some misunderstanding.

drishu: yes, I forgot

modules/oe_translation_cdt/config/install/field.field.oe_translation_request.cdt.request_status.yml
intelektron: I re-used the field defined in the remote module, for now I guess there is no need for dedicated status field, CDT is less complicated than ePoetry

drishu (approved):

PR #299, opened 2024-03-22
D8AGE-289 Create callback endpoints for CDT.

intelektron:
### Description

Two callback endpoints for updating translation requests and languages status.
The related change: https://github.com/openeuropa/cdt-client/pull/19

### Change log

- Added: Callback controller with tests


modules/oe_translation_cdt/src/CdtLanguageMapper.php
```diff
@@ -0,0 +1,44 @@
...
+   * @param string $langcode
+   *   The Drupal language code.
+   * @param \Drupal\oe_translation_cdt\TranslationRequestCdtInterface $translation_request
+   *   The translation request whose translator we need to check.
+   *
+   * @return string
+   *   The CDT language code.
+   */
+  public static function getCdtLanguageCode(string $langcode, TranslationRequestCdtInterface $translation_request): string {
+    // @todo Implement mapping configuration in the provider.
```
intelektron: To be implemented in D8AGE-290

modules/oe_translation_cdt/src/Controller/CallbackController.php
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * @param \Symfony\Component\HttpFoundation\Request $request
+   *   The request.
+   *
+   * @return bool
+   *   TRUE if the API key is valid, FALSE otherwise.
+   */
+  private function isApiKeyValid(Request $request): bool {
+    $request_api_key = $request->headers->get('apikey');
+    $site_api_key = Settings::get('cdt.api_key');
```
intelektron, line 33: This will have to be added to the requirements, but I would wait with it until we discuss D8AGE-394

modules/oe_translation_cdt/src/Controller/CallbackController.php
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   */
+  public function requestStatus(Request $request): Response {
+    if (!$this->isApiKeyValid($request)) {
+      throw new AccessDeniedHttpException('Invalid API key');
+    }
+    $request_status = CallbackSerializer::deserializeRequestStatus($request->getContent());
```
intelektron, line 89: This is implemented here: [src/Serializer/CallbackSerializer.php](https://github.com/openeuropa/cdt-client/pull/19/files#diff-c123044849c1d3b465c63d1b474febe569b1fd6d8027ceb1ef0bbe651822ed3c)

modules/oe_translation_cdt/src/Controller/CallbackController.php
```diff
@@ -0,0 +1,134 @@
...
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   */
+  public function jobStatus(Request $request): Response {
+    if (!$this->isApiKeyValid($request)) {
+      throw new AccessDeniedHttpException('Invalid API key');
+    }
+    $job_status = CallbackSerializer::deserializeJobStatus($request->getContent());
```
intelektron, line 120: This is implemented here: [src/Serializer/CallbackSerializer.php](https://github.com/openeuropa/cdt-client/pull/19/files#diff-c123044849c1d3b465c63d1b474febe569b1fd6d8027ceb1ef0bbe651822ed3c)

modules/oe_translation_cdt/src/Controller/CallbackController.php
```diff
@@ -0,0 +1,134 @@
...
+    if (!$this->isApiKeyValid($request)) {
+      throw new AccessDeniedHttpException('Invalid API key');
+    }
+    $request_status = CallbackSerializer::deserializeRequestStatus($request->getContent());
+    $translation_request = $this->loadTranslationRequest($request_status->getRequestIdentifier(), $request_status->getCorrelationId());
+    if (!$translation_request) {
+      throw new NotFoundHttpException('Translation request not found');
+    }
+    $translation_request->setRequestStatusFromCdt($request_status->getStatus());
+    if (!$translation_request->getCdtId()) {
```
intelektron, line 95: This condition is for the case when we get the Permanent ID based on the Correlation ID. This way we avoid calling the endpoint continuously to check if Permanent ID is available.

composer.json
```diff
@@ -29,6 +29,7 @@
         "drupal/paragraphs": "^1.13",
         "drupal/typed_link": "^2.0",
         "drush/drush": "^12",
+        "openeuropa/cdt-client": "dev-D8AGE-289",
```
intelektron, line 32: I added this as a repository/branch temporarily, until it is merged and put on packagist.org.
It is in `require-dev`, the same as `epoetry-client`.

modules/oe_translation_cdt/src/CdtLanguageMapper.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\oe_translation_cdt;
+
+/**
+ * Maps the Drupal language to the configured CDT language.
+ */
+class CdtLanguageMapper {
```
intelektron: I've created this class similar to the ePoetry module.

Now that I'm developing D8AGE-290, I've researched some best approaches in Entity <=> DTO conversion, and it seems that Symfony-based projects should define dedicated mapper classes (right now I'm working on `TranslationRequestMapper`, which assembles a DTO from TranslationRequest entity). 

So the idea is good, but maybe we should create a `Mapper` directory and put all related mapper classes there.

Just putting this out there for your consideration.

modules/oe_translation_cdt/src/Controller/CallbackController.php
```diff
@@ -0,0 +1,134 @@
...
+      ]);
+    }
+
+    if ($request = reset($requests)) {
+      assert($request instanceof TranslationRequestCdtInterface);
+      return $request;
+    }
+    else {
+      return NULL;
+    }
```
drishu:
```suggestion
    return NULL;
```

modules/oe_translation_cdt/src/CdtLanguageMapper.php
drishu: sounds good to me

drishu (changes requested):

drishu (approved):

PR #300, opened 2024-04-04
D8AGE-288: Create ContentFormatter for XML.

intelektron:
### Description

The ticket is about creating a serializer/deserializer to convert TranslationRequest to an XML file and import the translated content.

Previous implementations (epoetry, tmgmt_cdt) relied on twig and SimpleXML, I used the serialization component because it is flexible, and allows the two-way conversion.

### Change log

- Added: Serializer/deserializer, DTO classes, tests


modules/oe_translation_cdt/src/ContentFormatter/XmlFormatter.php
```diff
@@ -0,0 +1,184 @@
...
+    return TranslationSourceHelper::unflatten($translation_data);
+  }
+
+  /**
+   * Creates and returns the serializer on-demand.
+   *
+   * @return \Symfony\Component\Serializer\Serializer
+   *   The serializer.
+   */
+  protected function serializer(): Serializer {
```
intelektron, line 149: I have used this type of design pattern to avoid having to create the serializer if it is not needed at all. In this case, the `serialization' module is not required.

Previous modules used TWIG and SimpleXML for serialization, but the Symfony approach is much more testable and also allows data import.

modules/oe_translation_cdt/src/Model/Transaction.php
```diff
@@ -0,0 +1,207 @@
...
+namespace Drupal\oe_translation_cdt\Model;
+
+use Symfony\Component\Serializer\Attribute\SerializedPath;
+
+/**
+ * Defines the DTO class for transaction.
+ */
+class Transaction {
+
+  #[SerializedPath('[TransactionHeader][SenderDetails][@drupalVersion]')]
```
intelektron, line 14: In the current version of `drupal/coder`, this must go first, before the actual comment. The fix is on the way (https://www.drupal.org/project/coder/issues/3322615).

I used PHP attributes instead of annotations, because serializer annotations are deprecated and will be removed in SF7/D11.

modules/oe_translation_cdt/src/Model/Transaction.php
```diff
@@ -0,0 +1,207 @@
...
+
+  #[SerializedPath('[TransactionHeader][SenderDetails][ProducerDateTime]')]
+  /**
+   * The date of the file.
+   *
+   * @var \DateTimeInterface
+   */
+  protected \DateTimeInterface $producerDateTime;
+
+  #[SerializedPath('[TransactionIdentifier]')]
```
intelektron, line 34: I could possibly create a `NameConverter` class here to support the automatic conversion from `camelCase` to `PascalCase`, but the question is if it makes sense for few properties. Redefining the `SerializePath` works identically.

modules/oe_translation_cdt/tests/fixtures/cdt-source-file.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<Transaction xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="//static.cdt.europa.eu/webtranslations/schemas/Drupal-Translation-V8-2.xsd">
+  <TransactionHeader>
+    <SenderDetails drupalVersion="@drupal_version" moduleVersion="8.x-1.6">
```
intelektron: The parameters starting with `@` will be changed later by a simple `str_replace` inside a test.

I generate this file in the same way as `tmgmt_cdt` module does.

modules/oe_translation_cdt/xml/schema.xsd
```diff
@@ -0,0 +1,61 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" attributeFormDefault="unqualified" elementFormDefault="qualified">
```
intelektron, line 1: I've put this schema in the module to have a clear image of the XML format for later.

modules/oe_translation_cdt/src/ContentFormatter/XmlFormatter.php
```diff
@@ -0,0 +1,181 @@
...
+use Drupal\Component\Datetime\Time;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Render\RendererInterface;
+use Drupal\oe_translation\Entity\TranslationRequestInterface;
+use Drupal\oe_translation\TranslationSourceHelper;
+use Drupal\oe_translation_cdt\Model\Transaction;
+use Drupal\oe_translation_cdt\Model\TransactionItem;
+use Drupal\oe_translation_cdt\Model\TransactionItemField;
+use Symfony\Component\PropertyInfo\Extractor\PhpDocExtractor;
+use Symfony\Component\Serializer\Encoder\XmlEncoder;
```
drishu, line 17: the symfony serializer is a dependency of a dependency, we cannot assume it will always be present, we need to require dev this (see how it was done with symfony property-access), and put a small readme.md listing the modules requirements like in some of the other submodules

modules/oe_translation_cdt/src/ContentFormatter/XmlFormatter.php
```diff
@@ -0,0 +1,181 @@
...
+      $entity = $request->getContentEntity();
+      $transaction_item_fields[] = (new TransactionItemField())
+        ->setIndexType((string) $entity?->getEntityTypeId())
+        ->setInstanceId((string) $entity?->id())
+        ->setResourceName('][' . $key)
+        ->setResourceType(self::isHtml($field['#text']) ? 'html' : 'text')
+        ->setResourceLabel(implode(' / ', $field['#parent_label']))
+        ->setResourceMaxSize($field['#max_length'] ?? NULL)
+        ->setContent($field['#text']);
+      $character_count += mb_strlen(strip_tags($field['#text']));
```
drishu, line 71: In the email thread I saw that the rule for calculating the character count is "Calculated with all characters minus spaces.", neither mb_strlen nor strip_tags removes spaces

modules/oe_translation_cdt/tests/src/Kernel/XmlFormatterTest.php
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * Test the XML content formatter export.
+   */
+  public function testXmlContentExporter(): void {
+    $time_prophecy = $this->prophesize(Time::class);
+    $time_prophecy->getCurrentTime()->willReturn(1710935729);
+
+    $formatter = new XmlFormatter(
```
drishu: cant put my finger on it but it looks a bit strange, was expecting to see this called through the service, maybe give this a try `$this->container->set('datetime.time', $time_prophecy->reveal());`

modules/oe_translation_cdt/src/ContentFormatter/ContentFormatterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Parses the translated XML file.
+   *
+   * @param string $file
+   *   The XML string to import.
+   * @param \Drupal\oe_translation\Entity\TranslationRequestInterface $request
+   *   The translation request entity.
+   *
+   * @return mixed[]
```
drishu: `mixed[]` is an array of any type, isnt that the same as `array` but the later is less confusing?

modules/oe_translation_cdt/src/Model/Transaction.php
```diff
@@ -0,0 +1,207 @@
...
+    return $this->drupalVersion;
+  }
+
+  /**
+   * Sets the Drupal version.
+   *
+   * @param string $drupalVersion
+   *   The Drupal version.
+   */
+  public function setDrupalVersion(string $drupalVersion): self {
```
drishu: the convention is variable names are snake case $drupal_version and class attributes yes are lower camel case so $this->drupalVersion

drishu (changes requested):

modules/oe_translation_cdt/src/ContentFormatter/XmlFormatter.php
intelektron: That's true, but I was asking about the `volume` parameter in JSON request. It will be implemented in D8AGE-290.
The parameter in the XML `Is just informative. Total characters count for the job`

modules/oe_translation_cdt/tests/src/Kernel/XmlFormatterTest.php
intelektron: You're right, it made sense when I had more prophecies here, at the beginning of the ticket.

modules/oe_translation_cdt/src/ContentFormatter/ContentFormatterInterface.php
intelektron: I've added it while trying to be PHPStan-compatible, but you're right it is equivalent to `array`. I'll change it.

modules/oe_translation_cdt/src/Model/Transaction.php
intelektron: You are right, I got too comfortable with PHPStorm getter/setter generation. Will fix.
It is not reported by CodeSniffer because it is only [forbidden](https://www.drupal.org/docs/develop/standards/php/php-coding-standards#s-functions-and-variables) to mix cases in one file.

modules/oe_translation_cdt/src/ContentFormatter/XmlFormatter.php
drishu: ok, I though it was the same as volume

drishu (approved):

imanoleguskiza (approved):

PR #301, opened 2024-04-08
D8AGE-424: Create RequestMapper for CDT.

intelektron:
### Description

The responsibility of the mapper is to take the TranslationRequest entity and convert it to the Request Translation DTO. The conversion is only one-way, there is no need to convert the DTO to an entity.


modules/oe_translation_cdt/src/TranslationRequestCdt.php
```diff
@@ -56,14 +56,14 @@ public function updateTargetLanguageStatusFromCdt(string $langcode, string $cdt_
   /**
    * {@inheritdoc}
    */
-  public function getComments(): string {
+  public function getComments(): ?string {
```
intelektron, line 59: This change is necessary because comments are not required in the entity.

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Converts Drupal TranslationRequest entity to CDT library DTO.
+   *
+   * @param \Drupal\oe_translation_cdt\TranslationRequestCdtInterface $entity
+   *   The translation request entity.
+   *
+   * @return \OpenEuropa\CdtClient\Model\Request\Translation
+   *   The translation request DTO.
+   */
+  public function convertEntityToDto(ContentEntityInterface $entity): Translation {
```
drishu: There is a mismatch between what is in annotation and what is set as parameter type here.
Same for convertEntityToDto vs createSourceDocument, different param types even though they get the same entity object.

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestMapperTest.php
```diff
@@ -0,0 +1,144 @@
...
+      sprintf(
+        'translation_job_%s_request.xml',
+        $request->id()),
+      $file->getFileName()
+    );
+
+    $callbacks = $dto->getCallbacks();
+    foreach (['JOB_STATUS', 'REQUEST_STATUS'] as $key => $callback_type) {
+      $this->assertEquals('test_api_key', $callbacks[$key]->getClientApiKey());
+      $this->assertTrue(UrlHelper::isValid($callbacks[$key]->getCallbackBaseUrl()));
```
drishu, line 147: We could also check isExternal false to be sure it serves our drupal paths

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestMapperTest.php
```diff
@@ -0,0 +1,144 @@
...
+      $this->assertEquals($callback_type, $callbacks[$key]->getCallbackType());
```
drishu, line 151: I see a lot of positive testing, I think its not a bad idea to see some negative as well, for example what the output is if comments are not set (since they are optional), maybe check only with required fields first, then with all, and then maybe set some wrong values and check how it handles those

drishu (changes requested):

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
intelektron: This is because I planned to introduce a more general interface for DTO mappers. But maybe it doesn't make sense in this case, as we only have one mapper. I will change it.

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestMapperTest.php
intelektron: There's not much to test here, but I've added checking optional parameters and expecting exception when the entity is not complete. I also decided to check the volume counting.

drishu (approved):

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
```diff
@@ -0,0 +1,191 @@
...
+  protected function createSourceDocument(TranslationRequestCdtInterface $translation_request): SourceDocument {
+    $source_document = new SourceDocument();
+    $translation_jobs = $this->createTranslationJobs($translation_request);
+    $source_document->setTranslationJobs($translation_jobs);
+    $source_languages = [
+      LanguageCodeMapper::getCdtLanguageCode($translation_request->getSourceLanguageCode(), $translation_request),
+    ];
+    $source_document->setSourceLanguages($source_languages);
+    $source_document->setIsPrivate(FALSE);
+    $source_document->setOutputDocumentFormatCode('XM');
```
imanoleguskiza, line 81: I guess this should be XML?

imanoleguskiza (changes requested):

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
intelektron: Actually, it is intended, like [here](https://git.drupalcode.org/project/tmgmt_cdt/-/blob/8.x-1.x/src/Plugin/tmgmt/Translator/TmgmtCdtTranslator.php#L304). The `XML` throws a validation error.

imanoleguskiza (approved):

PR #302, opened 2024-04-09
D8AGE-425: Create API wrapper for CDT client.

intelektron:
### Description

The goal of the ticket was to create an abstraction layer to wrap the CDT client and set the bearer token when needed.

My first idea was to create a `CdtClient` class that would serve the `ApiClient` object from the cdt-client library, but then it turned out that this was not a good pattern as it broke SRP and prevented me from using PHPUnit prophecies properly.

Finally, I added the ServiceProvider to serve the `ApiClient` directly from the library. This way we can easily create a dummy test service from the prophecy. There's an additional `CdtApiAuthenticator` service that adds the authorization token when needed. It reads the token from the state and checks the connection.

Please let me know what you think of this solution.


modules/oe_translation_cdt/src/Api/CdtApiAuthenticator.php
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    $expiry = $this->state->get('cdt.token_expiry_date');
+    $is_expired = $expiry && $expiry < $this->time->getCurrentTime();
+
+    /** @var \OpenEuropa\CdtClient\Model\Response\Token|false $state_token */
+    $state_token = unserialize((string) $this->state->get('cdt.token'), [
+      'allowed_classes' => [Token::class],
+    ]);
+
```
drishu: Lets check that the token is of the correct class

modules/oe_translation_cdt/src/OeTranslationCdtServiceProvider.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Registers a CDT Client service with custom config.
+ */
+class OeTranslationCdtServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function register(ContainerBuilder $container): void {
```
drishu, line 19: I was expecting a single service class that handles instantiation and authentication, but I dont mind this approach

modules/oe_translation_cdt/tests/src/Kernel/CdtApiAuthenticatorTest.php
```diff
@@ -0,0 +1,160 @@
...
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'oe_translation_cdt',
+  ];
+
+  /**
+   * The API client prophecy.
+   */
+  private ApiClientInterface|ObjectProphecy $apiProphecy;
```
drishu: this is a bit confusing for me and my editor, this returns Prophecy\Prophecy\ObjectProphecy

drishu (changes requested):

modules/oe_translation_cdt/tests/src/Kernel/CdtApiAuthenticatorTest.php
intelektron: I took this approach [from core](https://git.drupalcode.org/project/drupal/-/blob/10.3.x/core/modules/field/tests/src/Unit/Plugin/migrate/process/ProcessFieldTest.php?ref_type=heads#L34). Unfortunately, PHPStorm doesn't like Prophecy by default. It requires a [plugin](https://plugins.jetbrains.com/plugin/9674-phpunit-enhancement) to properly understand it.

modules/oe_translation_cdt/src/OeTranslationCdtServiceProvider.php
intelektron: Creating a wrapper class with a `->getClient()` factory was my first thought.

But then I started writing tests, and it turned out that having the client as a standalone service simplifies the use of PHPUnit prophecies. I can just create a dummy object and point it to the container. All other services use it from that moment on.

Having an additional layer of logic can complicate things a bit. For example, if I have an `ApiWrapper` service that creates an `ApiClient` and authenticates it, I would have to prophesize its `getClient()` and set the dummy `ApiWrapper` service to the container. If I do this, I won't be able to test the authentication methods of the service because they will also become dummy. According to the Prophecy docs, we should separate classes so that they are testable (classic SRP rule).

The service provider is necessary because I need to read the static settings and pass them to the constructor. I can't do this using the services.yml file.

**However**, after having some reading, I could make few improvements to this code 🙂

1. I can change the `ApiAuthenticator` class to a classic `ApiWrapper` class that provides the `getClient()` method that gets and authenticates the client. The `ApiClient` is passed to the service via dependency injection.
2. The `ApiClient` will be defined in `*.services.yml` as `public: false`, so it can only be a dependency. It will make use of container parameters to set up the constructor variables.

```yml
services:
  oe_translation_cdt.api_wrapper:
    class: Drupal\oe_translation_cdt\Api\CdtApiWrapper
    arguments: ['@oe_translation_cdt.api_client', '@http_client', '@state', '@datetime.time']
    
  oe_translation_cdt.api_client:
    class: OpenEuropa\CdtClient\ApiClient
    public: false
    arguments:
      - @ http_client
      - @ psr17.server_request_factory
      - @ psr17.stream_factory
      - @ oe_translation_cdt.psr17.uri_factory
      - %cdt.client_configuration%
```

3. I define a new compiler pass, and set the `cdt.client_configuration` parameter to the container. This is how core does it.

modules/oe_translation_cdt/oe_translation_cdt.services.yml
```diff
@@ -0,0 +1,18 @@
+services:
+  oe_translation_cdt.api_client:
```
intelektron, line 2: Unfortunately, I couldn't set it to private. It blocks the service replacement in tests. Forcing it to be public would require too many overrides.

modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
```
drishu, line 2: Let's add `declare(strict_types=1);` where missing

modules/oe_translation_cdt/src/OeTranslationCdtServiceProvider.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Runs the custom compiler pass.
+ */
+class OeTranslationCdtServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function register(ContainerBuilder $container): void {
+    $container->addCompilerPass(new CdtParametersCompilerPass());
```
drishu, line 20: Nice approach 👍

drishu (changes requested):

drishu (approved):

imanoleguskiza (approved):

PR #304, opened 2024-04-19
EWPP-4243: Fixing multivalue translation removals.

upchuk:
Fixes the case in which a multivalue field has a value removed in the source and we sync the translation - the (correct) value should also be removed in the translation.

It fixes also a bug in the form for creating a field using translation multivalue.

### DIFF
--- a/modules/oe_translation_local/modules/oe_translation_multivalue/tests/src/Functional/MultivalueTranslationsTest.php
+++ b/modules/oe_translation_local/modules/oe_translation_multivalue/tests/src/Functional/MultivalueTranslationsTest.php
@@ -176,18 +176,335 @@ class MultivalueTranslationsTest extends TranslationTestBase {
    * @SuppressWarnings(PHPMD.NPathComplexity)
    */
   public function testMultivalueFieldTranslationsApi(): void {
-    // Enable the multivalue translation on our fields.
-    $field_names = [
-      'field_textfield',
-      'field_address',
-      'field_description_list',
-      'field_link',
-      'field_link_description',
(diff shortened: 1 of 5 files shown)


modules/oe_translation_local/modules/oe_translation_multivalue/tests/src/Functional/MultivalueTranslationsTest.php
```diff
@@ -405,6 +265,70 @@ public function testMultivalueFieldTranslationsApi(): void {
...
+    $paragraph->save();
+    foreach ($this->getMultivalueFieldNames() as $field_name) {
+      $values = $node->get($field_name)->getValue();
+      unset($values[0]);
+      $node->set($field_name, $values);
+      $node->set('field_paragraphs', [
+        'target_id' => $paragraph->id(),
+        'target_revision_id' => $paragraph->getRevisionId(),
+      ]);
+      $node->setNewRevision();
```
sergepavle: I didn't get why we are updating same field value in each iteration. 
```suggestion
    $paragraph->save();
    $node->set('field_paragraphs', [
      'target_id' => $paragraph->id(),
      'target_revision_id' => $paragraph->getRevisionId(),
    ]);
    foreach ($this->getMultivalueFieldNames() as $field_name) {
      $values = $node->get($field_name)->getValue();
      unset($values[0]);
      $node->set($field_name, $values);
      $node->setNewRevision();
```

sergepavle (dismissed):

hernani (dismissed):

PR #305, opened 2024-04-25
D8AGE-426: Create the mocking module.

intelektron:
### Description

The PR contains a new `oe_translation_cdt_mock` module for mocking the CDT requests. It uses the `http_request_mock` module to intercept requests sent to the CDT endpoint.


docker-compose.yml
```diff
@@ -26,18 +26,18 @@ services:
...
-      CDT_VALIDATE_API_ENDPOINT: ""
-      CDT_REQUESTS_API_ENDPOINT: ""
-      CDT_IDENTIFIER_API_ENDPOINT: ""
-      CDT_STATUS_API_ENDPOINT: ""
-      CDT_FILE_API_ENDPOINT: ""
-      CDT_CLIENT: ""
-      CDT_USERNAME: ""
-      CDT_PASSWORD: ""
-      CDT_API_KEY: ""
+      CDT_TOKEN_API_ENDPOINT: "https://example.com/api/token"
```
intelektron, line 29: I had to add these URLs, as the mocking module uses them too.
This is the downside of using static Settings, they can't be overridden by the mocking module in a clean way.

.drone.yml
```diff
@@ -24,18 +24,18 @@ services:
...
-      - CDT_VALIDATE_API_ENDPOINT=""
-      - CDT_REQUESTS_API_ENDPOINT=""
-      - CDT_IDENTIFIER_API_ENDPOINT=""
-      - CDT_STATUS_API_ENDPOINT=""
-      - CDT_FILE_API_ENDPOINT=""
-      - CDT_CLIENT=""
-      - CDT_USERNAME=""
-      - CDT_PASSWORD=""
-      - CDT_API_KEY=""
+      - CDT_TOKEN_API_ENDPOINT=https://example.com/api/token
```
intelektron, line 27: Please see the comment in `docker-compose.yml`

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/responses/json/reference_data_response_200.json
```diff
@@ -0,0 +1,322 @@
...
+      "description": "Send"
+    },
+    {
+      "code": "SendAsDraft",
+      "description": "Send as draft"
+    }
+  ],
+  "contacts": [
+    {
+      "email": "test@example.com",
```
intelektron, line 312: This value is sanitized

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/responses/json/status_response_200.json
```diff
@@ -0,0 +1,149 @@
...
+  "requestIdentifier": "2024/000123",
+  "status": "COMP",
+  "sourceLanguages": [
+    "EN"
+  ],
+  "targetLanguages": [
+    "DE",
+    "ET"
+  ],
+  "creationDate": "2024-03-31T07:45:46.3595527",
```
intelektron, line 11: The values here come from a sample response. I am changing them in the MockService if needed.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/FileApi.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+declare(strict_types=1);
```
intelektron, line 3: I prepared this mock to work with the non-deprecated version of the API. We will have to finalize the https://github.com/openeuropa/cdt-client/pull/18 to make use of it.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/FileApi.php
```diff
@@ -0,0 +1,110 @@
...
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('extension.list.module'),
+      $container->get('entity_type.manager'),
+      $container->get('logger.factory'),
+      $container->get('oe_translation_cdt.xml_formatter')
```
intelektron, line 69: I've overridden the constructor to inject the additional `xml_formatter` service.
Please let me know what you think of it. Maybe it should be in the base class,

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/FileApi.php
```diff
@@ -0,0 +1,110 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getEndpointUrl(): string {
+    // The file URL is not present in the static settings.
+    // It is defined by the Status API response.
+    return 'https://example.com/api/files/:language/:id';
```
intelektron, line 79: This URL is not defined in the static settings, it is returned by the `Status` calls.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/FileApi.php
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
```
intelektron, line 85: The mock works like this:

- The mock file URL contains the language and the translation request ID (the production URL contains only an UUID).
- We take all the translatable data stored in the translation request, and generate a new XML file with modified text.
- All the field content is changed from `Foo` to ex. `fr translation of Foo`.
- The translated XML file has the same format as the source XML file, this is why it is possible to reuse `XmlFormatter` service.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/IdentifierApi.php
```diff
@@ -0,0 +1,39 @@
...
+   * {@inheritdoc}
+   */
+  protected function getEndpointUrl(): string {
+    return Settings::get('cdt.identifier_api_endpoint');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
```
intelektron, line 33: The mocked correlation ID is identical to the translation request entity ID (`123`).
The mocked permanent ID is the current year, followed by the translation request entity ID (`2024/123`).

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/RequestsApi.php
```diff
@@ -0,0 +1,42 @@
...
+   */
+  protected function getEndpointUrl(): string {
+    return Settings::get('cdt.requests_api_endpoint');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
+    // Make the correlation ID equal to the translation request ID.
```
intelektron, line 34: This mock fakes the acceptance unconditionally and returns the correlation ID (equal to the entity ID).

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/ServiceMockBase.php
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Determines if the endpoint needs authorization.
+   *
+   * Override it to FALSE for public endpoints.
+   *
+   * @return bool
+   *   TRUE if the authorization token is required, FALSE otherwise.
+   */
+  protected function needsToken(): bool {
+    return TRUE;
```
intelektron, line 93: This is needed for the Token API that a public endpoint

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/ServiceMockBase.php
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Gets the request parameters from the request path.
+   *
+   * @param \Psr\Http\Message\RequestInterface $request
+   *   The request.
+   *
+   * @return array
+   *   The request parameters in the associative array.
+   */
+  protected function getRequestParameters(RequestInterface $request): array {
```
intelektron, line 105: Some of the endpoints contain parameters in the path (and not the query string).
Like: https://example.com/api/status/:requestyear/:requestnumber

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/StatusApi.php
```diff
@@ -0,0 +1,83 @@
...
+    $parameters = $this->getRequestParameters($request);
+    $response = json_decode($this->getResponseFromFile('status_response_200.json'), TRUE);
+    /** @var \Drupal\oe_translation_cdt\TranslationRequestCdtInterface|null $entity */
+    $entity = $this->entityTypeManager->getStorage('oe_translation_request')->load($parameters['requestnumber']);
+    if (!$entity) {
+      $this->log('400: Failed to get the mock status. The translation request does not exist.', $request);
+      return new Response(400, [], $this->getResponseFromFile('status_response_400.json'));
+    }
+
+    // Change only the important parameters.
```
intelektron, line 46: I did not change any other parameters of this request, as Drupal does not check it for now. This may change in [D8AGE-291](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/D8AGE-291), I will modify the mock accordingly.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/StatusApi.php
```diff
@@ -0,0 +1,83 @@
...
+  protected function getEndpointUrl(): string {
+    return Settings::get('cdt.status_api_endpoint');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
+    // The entity uses the same identifier as the mocked CDT request.
+    $parameters = $this->getRequestParameters($request);
```
intelektron, line 37: I decided not to keep any additional data structures that would be used for the mock.
This mock fetches the translation request entity and constructs a matching JSON response.
This response should not actually update the request data.

To fake the translation process moving forward, we will use the fake callbacks.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/StatusApi.php
```diff
@@ -0,0 +1,83 @@
...
+
+    // Change only the important parameters.
+    $response['status'] = match($entity->getRequestStatus()) {
+      TranslationRequestRemoteInterface::STATUS_REQUEST_TRANSLATED => 'COMP',
+      TranslationRequestRemoteInterface::STATUS_REQUEST_FAILED_FINISHED => 'CANC',
+      TranslationRequestRemoteInterface::STATUS_REQUEST_REQUESTED => 'INPR',
+      default => 'UNDE',
+    };
+
+    // Add source language and translated files.
```
intelektron, line 54: If any of the languages is already finished, it will be added as a target file to the response.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/StatusApi.php
```diff
@@ -0,0 +1,83 @@
...
+      if ($language->getStatus() != TranslationRequestRemoteInterface::STATUS_LANGUAGE_REQUESTED) {
+        $response['targetFiles'][] = [
+          'sourceLanguage' => $cdt_source_language,
+          'targetLanguage' => $cdt_target_language,
+          'sourceDocument' => 'test.xml',
+          'fileName' => 'test.xml',
+          'isPrivate' => FALSE,
+          '_links' => [
+            'files' => [
+              'href' => sprintf('https://example.com/api/files/%s/%s', $cdt_target_language, $entity->id()),
```
intelektron, line 71: This URL will be matched by the `FileApi.php`.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/TokenApi.php
```diff
@@ -0,0 +1,45 @@
...
+   */
+  protected function needsToken(): bool {
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
+    $this->log('200: Returning the mocked authorization token.', $request);
```
intelektron, line 41: Authorize the connection unconditionally, regardless of the username and password.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/src/Plugin/ServiceMock/ValidateApi.php
```diff
@@ -0,0 +1,38 @@
...
+   */
+  protected function getEndpointUrl(): string {
+    return Settings::get('cdt.validate_api_endpoint');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointResponse(RequestInterface $request): ResponseInterface {
+    $this->log('200: Successfully validating the translation request.', $request);
```
intelektron, line 34: Validate the request unconditionally, regardless of the content.
Please let me know if this should include validation logic.

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/tests/src/Unit/ServiceMockBaseTest.php
```diff
@@ -0,0 +1,102 @@
+<?php
```
intelektron, line 1: I know that testing the mock may be too much, but I wanted to make sure that the utility functions are working fine.

drishu (approved):

modules/oe_translation_cdt/modules/oe_translation_cdt_mock/responses/json/status_response_200.json
imanoleguskiza: I am not sure 100% this is the best way to go to be honest, it feels weird to not populate this values based on the actual request being sent. Ultimately I think the idea is to add mock actions as well just like the epoetry does and in that case we will need to also send requests to our endpoint and the information should match what is stored in the request. In any case, we don't have that use case for now so we could merge as is and then handle it differently when the need arises.

imanoleguskiza (approved):

PR #306, opened 2024-04-25
D8AGE-290: Add translation request remote provider.

intelektron:
### Description

The goal of the ticket was to add a new `Cdt.php` remote provider, that would allow us to create and send translation requests. The provider defines a form (connected to remote ReferenceData) and uses some components (`XmlFormatter`, `ApiWrapper`) from previous tickets.

### DIFF
--- a/modules/oe_translation_cdt/oe_translation_cdt.module
+++ b/modules/oe_translation_cdt/oe_translation_cdt.module
@@ -7,6 +7,8 @@
 
 declare(strict_types=1);
 
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\oe_translation_cdt\TranslationRequestCdt;
 
 /**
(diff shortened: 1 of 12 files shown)


modules/oe_translation_cdt/oe_translation_cdt.module
```diff
@@ -17,3 +19,44 @@ function oe_translation_cdt_entity_bundle_info_alter(array &$bundles): void {
     $bundles['oe_translation_request']['cdt']['class'] = TranslationRequestCdt::class;
   }
 }
+
```
intelektron, line 22: This is actually a part of D8AGE-341 copied from ePoetry, but I've included it here to have a complete functional test. Without this, the list of translations is broken.

modules/oe_translation_cdt/src/Plugin/RemoteTranslationProvider/Cdt.php
```diff
@@ -0,0 +1,342 @@
...
+      return isset($default_configuration[$key]);
+    }, ARRAY_FILTER_USE_KEY);
+    $this->configuration = $configuration + $default_configuration;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildConfigurationForm(array $form, FormStateInterface $form_state): array {
+    $form['language_mapping'] = [
```
intelektron, line 135: I did not include auto-accept and auto-synchronize here, as it is impossible to test on this stage.

modules/oe_translation_cdt/src/Plugin/RemoteTranslationProvider/Cdt.php
```diff
@@ -0,0 +1,342 @@
...
+        'deliver_to' => $reference_data_object->getContacts(),
+        'contact_usernames' => $reference_data_object->getContacts(),
+      ];
+    }
+    catch (InvalidStatusCodeException | CdtConnectionException $exception) {
+      $this->loggerFactory->get('oe_translation_cdt')->error($exception->getMessage());
+      $form['message'] = [
+        '#markup' => $this->t('There was a problem retrieving the reference data from CDT.'),
+      ];
+      return $form;
```
intelektron, line 178: The "Send" button from the parent form is still active at this point, but the validation won't pass anyway.

modules/oe_translation_cdt/tests/src/Kernel/CallbackControllerTest.php
```diff
@@ -113,10 +113,11 @@ public function testRequestStatusCallback(): void {
     $translation_request_with_id = TranslationRequest::create([
       'bundle' => 'cdt',
       'cdt_id' => '2024/12345a',
-      'cdt_status' => TranslationRequestRemoteInterface::STATUS_REQUEST_REQUESTED,
```
intelektron, line 116: This test requires an update after modifying the fields of the translation request entity. At least one language has to be defined here, otherwise the request will be considered as finished.

drishu (approved):

imanoleguskiza (approved):

PR #310, opened 2024-05-23
D8AGE-341: Create dashboard for CDT translations.

intelektron:
### Description

Provide a CDT Requests page with an overview of all requests similar to ePoetry.


modules/oe_translation_cdt/src/Plugin/views/filter/TargetLanguagesFilter.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
```
intelektron, line 3: A very similar filter exists inside ePoetry, however, moving it out to `oe_translation_remote` will require writing some updates (for IDs that will have to be changed), and I am not sure if it is worth it.

modules/oe_translation_cdt/tests/src/FunctionalJavascript/TranslationProviderTest.php
```diff
@@ -279,6 +279,11 @@ public function testCdtSingleTranslationFlow(): void {
       ],
     ];
     $this->assertLogMessagesTable($expected_logs);
+
+    // Check the translation dashboard.
```
intelektron, line 258: I've only added this to check for possible errors on this page. Custom plugins are covered by kernel tests, and the view itself is rather generic.

modules/oe_translation_cdt/src/Access/CdtAccessCheck.php
```diff
@@ -0,0 +1,57 @@
...
+use Drupal\Core\Access\AccessResultInterface;
+use Drupal\Core\Cache\CacheableMetadata;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Session\AccountInterface;
+
+/**
+ * Checks access to the CDT translations.
+ */
+class CdtAccessCheck implements AccessInterface {
```
intelektron, line 17: The code is similar to ePoetry, but I moved it to a separate service to provide checkers for other routes.
This way, we have a centralized access method, covered by kernel tests.

modules/oe_translation_cdt/tests/src/FunctionalJavascript/TranslationProviderTest.php
```diff
@@ -279,6 +279,11 @@ public function testCdtSingleTranslationFlow(): void {
...
+    $this->drupalGet('admin/content/cdt-translation-requests');
+    $this->assertSession()->pageTextContains($cdt_id);
+    $this->assertSession()->pageTextContains('Finished');
```
drishu, line 262: We should still check that users with different permissions have or not access for two reasons, first there is a views plugin that is not covered by testing, second the underlying access check for the view might change and we would not catch any issues. I think its better to test here the basic access, this way no matter what we do with the plugins we can always rely that the correct user has access.

drishu (changes requested):

drishu (approved):

modules/oe_translation_cdt/src/Plugin/views/field/TargetLanguagesWithTooltip.php
```diff
@@ -0,0 +1,143 @@
...
+    // Count the number of translated languages.
+    $translated_count = 0;
+    foreach ($target_languages as $language_with_status) {
+      if (isset($request->getTranslatedData()[$language_with_status->getLangcode()])) {
+        $translated_count++;
+      }
+    }
+
+    $elements = [
+      'tooltip' => [
```
imanoleguskiza: Not sure why, but when trying locally it is not loading the css library so it doesn't actually display as a tooltip

imanoleguskiza (changes requested):

modules/oe_translation_cdt/src/Plugin/views/field/TargetLanguagesWithTooltip.php
intelektron: At first, I suspected that ePoetry pulls this library, but it is managed directly by `oe_translation`.

After few installs, the tooltip disappeared for me too :(

The oe_translation.module defines the [theme](https://github.com/openeuropa/oe_translation/blob/ba8f0504dde4b57acb129b3fecab4248f60c4c54/oe_translation.module#L39) and adds the tooltop library in the [preprocess function](https://github.com/openeuropa/oe_translation/blob/ba8f0504dde4b57acb129b3fecab4248f60c4c54/oe_translation.module#L348). It looks pretty straightforward. I am still checking it.

intelektron: It took me some time to track this, it turned out that the `renderPlain` method was dropping the asserts. I've changed it to `render` in https://github.com/openeuropa/oe_translation/pull/310/commits/90ac5ce8bfaede6341dfdb31c88d31b88695dae2 and then adjusted the test in https://github.com/openeuropa/oe_translation/pull/310/commits/5fa7dc383c471d038c9bd353ce7c4fdb15a485eb.

imanoleguskiza (approved):

drishu (approved):

PR #313, opened 2024-05-29
EWPP-3783: Logging incorrect incoming epoetry request translation files.

upchuk:
In case the translators send the file of another translation request, we need to log this and return a proper error.

### DIFF
--- a/modules/oe_translation_epoetry/src/ContentFormatter/HtmlFormatter.php
+++ b/modules/oe_translation_epoetry/src/ContentFormatter/HtmlFormatter.php
@@ -91,6 +91,16 @@ class HtmlFormatter implements ContentFormatterInterface {
       return [];
     }
 
+    $assets = $xml->xpath("//div[@class='asset']");
+    if (!$assets) {
+      return [];
+    }
+
+    $asset_id = (string) $assets[0]->attributes()['id'];
+    if ($asset_id !== 'item-' . $request->id()) {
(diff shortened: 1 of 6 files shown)


modules/oe_translation_epoetry/src/ContentFormatter/HtmlFormatter.php
```diff
@@ -91,6 +91,16 @@ public function import(string $file, TranslationRequestInterface $request): arra
...
     }
 
+    $assets = $xml->xpath("//div[@class='asset']");
+    if (!$assets) {
+      return [];
+    }
+
+    $asset_id = (string) $assets[0]->attributes()['id'];
+    if ($asset_id !== 'item-' . $request->id()) {
+      throw new \Exception('The translation request file does not match the translation request.');
```
intelektron, line 101: Is it ok to use the global \Exception case in this case? It would suggest that all other exceptions (like type errors or other internal stuff) are considered as `The ePoetry notification did not provide a valid translation` by the logger. This may be misleading while investigating issues.

upchuk: Well, this exception is thrown only if the request IDs dont match, not for any other reasons, so the message is returned only in this case.

intelektron (approved):

imanoleguskiza (approved):

PR #320, opened 2024-07-01
D8AGE-542: Fix handling of container parameters.

intelektron:
### Description

Fix handling of container parameters. When we add them as an array, Drupal replaces `%` characters with `%%`, and it does this twice in `OptimizedPhpArrayDumper.php`. I'm not sure if this is a bug, but the password `ab%c` becomes `ab%%%%c`.

### DIFF
--- a/modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
+++ b/modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
@@ -9,7 +9,7 @@ use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
- * Registers the cdt.client_configuration container parameter.
+ * Registers the container parameters used by the CDT.
  */
 class CdtParametersCompilerPass implements CompilerPassInterface {
 
@@ -20,6 +20,7 @@ class CdtParametersCompilerPass implements CompilerPassInterface {
     // The container parameters can't be added as a single array.
     // Array values don't support the "%" character, that may
     // appear in the password or in the URL.
+    // @see https://www.drupal.org/project/drupal/issues/3458344
     $container->setParameter('cdt.base_api_url', Settings::get('cdt.base_api_url'));
     $container->setParameter('cdt.username', Settings::get('cdt.username'));
     $container->setParameter('cdt.password', Settings::get('cdt.password'));
--- a/modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
+++ b/modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
@@ -17,12 +17,13 @@ class CdtParametersCompilerPass implements CompilerPassInterface {
    * {@inheritdoc}
    */
   public function process(ContainerBuilder $container): void {
-    $container->setParameter('cdt.client_configuration', [
-      'apiBaseUrl' => Settings::get('cdt.base_api_url'),
-      'username' => Settings::get('cdt.username'),
-      'password' => Settings::get('cdt.password'),
-      'client' => Settings::get('cdt.client'),
-    ]);
+    // The container parameters can't be added as a single array.
+    // Array values don't support the "%" character, that may
+    // appear in the password or in the URL.
+    $container->setParameter('cdt.base_api_url', Settings::get('cdt.base_api_url'));
+    $container->setParameter('cdt.username', Settings::get('cdt.username'));
+    $container->setParameter('cdt.password', Settings::get('cdt.password'));
+    $container->setParameter('cdt.client', Settings::get('cdt.client'));
   }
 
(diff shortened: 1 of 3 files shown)


gorsky87 (approved):

modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
```diff
@@ -9,20 +9,22 @@
...
-      'password' => Settings::get('cdt.password'),
-      'client' => Settings::get('cdt.client'),
-    ]);
+    // The container parameters can't be added as a single array.
+    // Array values don't support the "%" character, that may
+    // appear in the password or in the URL.
+    // @see https://www.drupal.org/project/drupal/issues/3458344
+    $container->setParameter('cdt.base_api_url', Settings::get('cdt.base_api_url'));
+    $container->setParameter('cdt.username', Settings::get('cdt.username'));
+    $container->setParameter('cdt.password', Settings::get('cdt.password'));
```
drishu: when flushing the caches I get `The parameter "cdt.password" has a dependency on a non-existent parameter "ssword". Did you mean this: "cdt.password"?`
Symfony treats this `%` as a special character in the ParameterBag, so we need to double it, see ParameterBagL234, I think a simple `str_replace("%", "%%", Settings::get('cdt.password'));` here will do the trick

drishu (changes requested):

modules/oe_translation_cdt/src/Compiler/CdtParametersCompilerPass.php
intelektron: Very weird, I will doublecheck!
The problem is that Drupal tries to do this `str_replace` by itself, so it may result with `%%%%`

drishu: I still get the error, reinstalled everything from scratch, even re-pulled the docker containers after deleting them
This is the config with issues:
```
      CDT_BASE_API_URL: "https://example.com/api"
      CDT_CLIENT: "test_client"
      CDT_USERNAME: "test_username"
      CDT_PASSWORD: "tes#t_pa%ssw!or%%d"
      CDT_API_KEY: "test_apikey"
```
this is the command that generates the error
`./vendor/bin/drush en -y oe_translation_cdt_mock`

Steps to repro:
- add above config to the web image
- disable xdebug or use CI image
- composer install this branch
- run drupal:site-install
- run ./vendor/bin/drush en -y oe_translation_cdt_mock

intelektron: Indeed, there was another escaping issue here. I've just added `str_replace`, and made an additional test to be 100% sure

modules/oe_translation_cdt/tests/src/Kernel/CdtParametersCompilerPassTest.php
```diff
@@ -0,0 +1,73 @@
...
+      /**
+       * Constructs the parameter receiver object.
+       */
+      public function __construct(public array $configuration = []) {
+      }
+
+    };
+
+    // Create a new container and register the compiler pass.
+    $container = new ContainerBuilder();
```
intelektron, line 53: I didn't find any other way to recompile the container based on new settings. This is approach preferred in core tests.

modules/oe_translation_cdt/tests/src/Kernel/CdtParametersCompilerPassTest.php
```diff
@@ -0,0 +1,73 @@
...
+      'cdt.base_api_url' => '%url%',
+      'cdt.username' => '%username',
+      'cdt.client' => '%%client%%',
+      'cdt.password' => 'test_%%password%',
+      'cdt.api_key' => 'test_%api_key',
+    ];
+    new Settings($settings);
+
+    // Define a simple inline class that will receive the parameters.
+    $receiver_object = new class {
```
intelektron, line 42: I didn't use the original `ApiClient` because I'd have to make its properties available via the reflection. And this would be hard to maintain if anything changed on the cdt-client side. This minimalistic approach seems reasonable.

modules/oe_translation_cdt/tests/src/Kernel/CdtParametersCompilerPassTest.php
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Tests that the CDT configuration is passed properly to the container.
+   */
+  public function testCdtCompilerPass(): void {
+    // Define strings with "%" characters, that need to be escaped.
+    $settings = [
+      'cdt.base_api_url' => '%url%',
+      'cdt.username' => '%username',
+      'cdt.client' => '%%client%%',
+      'cdt.password' => 'test_%%password%',
```
drishu, line 36: maybe add some more querky characters !@#$%^ so on

drishu (changes requested):

PR #322, opened 2024-07-04
D8AGE-581: Make priority update optional.

intelektron:
### Description

The update of the priority field should be optional, it should happen only if the translation job was created.

### DIFF
--- a/modules/oe_translation_cdt/src/TranslationRequestUpdater.php
+++ b/modules/oe_translation_cdt/src/TranslationRequestUpdater.php
@@ -41,22 +41,23 @@ final class TranslationRequestUpdater implements TranslationRequestUpdaterInterf
    * {@inheritdoc}
    */
   public function updateFromTranslationResponse(TranslationRequestCdtInterface $translation_request, Translation $translation_response, ReferenceData $reference_data): bool {
-    $languages = [];
-    foreach ($translation_response->getJobSummary() as $job) {
-      $drupal_langcode = LanguageCodeMapper::getDrupalLanguageCode($job->getTargetLanguage(), $translation_request);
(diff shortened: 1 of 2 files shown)


modules/oe_translation_cdt/src/TranslationRequestUpdater.php
```diff
@@ -41,23 +41,24 @@ public function updateFromJobStatus(TranslationRequestCdtInterface $translation_
    * {@inheritdoc}
    */
   public function updateFromTranslationResponse(TranslationRequestCdtInterface $translation_request, Translation $translation_response, ReferenceData $reference_data): bool {
```
intelektron, line 43: The function was changed a bit to satisfy phpmd's Cyclomatic Complexity.

javitan (approved):

gorsky87 (approved):

modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestUpdaterTest.php
```diff
@@ -139,9 +139,9 @@ public function testUpdateFromTranslationResponse(): void {
       'Manually updated the status.' .
       'Updated request_status field from Requested to Translated.' .
       'The following languages are updated: fr (Requested =&gt; Review).' .
-      'Updated priority field from PRIO1 to PRIO2.' .
       'Updated comments field to TEST.' .
       'Updated phone_number field from 999999999 to 123456789.' .
+      'Updated priority field from PRIO1 to PRIO2.' .
```
drishu, line 144: we can cover this a bit better with testing by not setting the job summary. Perhaps a translation response with only minimal or required fields, assert, new response with all fields, assert

PR #323, opened 2024-07-17
D8AGE-581: Enhance the Request Updater test to check optional values.

intelektron:
## D8AGE-581

The follow-up of https://github.com/openeuropa/oe_translation/pull/322

### DIFF
--- a/modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
+++ b/modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_translation_cdt;
+
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\oe_translation\Entity\TranslationRequest;
+use Drupal\oe_translation_cdt\TranslationRequestCdtInterface;
+use Drupal\oe_translation_remote\TranslationRequestRemoteInterface;
+
+/**
+ * Helpers for testing CDT.
+ */
+trait CdtTranslationTestTrait {
+
+  /**
+   * Gets the common translation request parameters.
+   *
+   * @return array
+   *   The common parameters of CDT translation request.
+   */
+  private function getCommonTranslationRequestData() {
+    return [
+      'bundle' => 'cdt',
+      'translator_provider' => 'cdt',
+      'correlation_id' => '12345',
+      'request_status' => TranslationRequestRemoteInterface::STATUS_REQUEST_REQUESTED,
+      'comments' => 'COMMENT1',
+      'confidentiality' => 'test_confidentiality',
+      'contact_usernames' => [
+        'TEST1',
+      ],
+      'deliver_to' => [
+        'TEST2',
+      ],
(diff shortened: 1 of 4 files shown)


modules/oe_translation_cdt/tests/src/Kernel/TranslationRequestUpdaterTest.php
```diff
@@ -221,6 +228,7 @@ protected function createTranslationRequest(): TranslationRequestCdtInterface {
    *   The expected log messages. There may be wildcards "*" inside.
    */
   protected function assertTranslationRequestLog(TranslationRequestCdtInterface $request, array $logs): void {
+    $this->assertEquals(count($logs), count($request->getLogMessages()), 'The log size doesn\'t match.');
```
drishu: assertSameSize

drishu (changes requested):
We have two createTranslationRequest we should sync their usage into one function and put it into a trait

modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+trait CdtTranslationTestTrait {
+
+  /**
+   * Gets the common translation request parameters.
+   *
+   * @return array
+   *   The common parameters of CDT translation request.
+   */
+  private function getCommonTranslationRequestData() {
```
drishu: add return type hinting, and I think we can make it static

modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_translation_cdt;
```
drishu: this should go into a "Traits" folder

modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
```diff
@@ -0,0 +1,72 @@
...
+   *   The data to create the translation request.
+   * @param \Drupal\Core\Entity\ContentEntityInterface|null $entity
+   *   The content entity to attach to the translation request.
+   * @param array $languages
+   *   The target languages.
+   *
+   * @return \Drupal\oe_translation_cdt\TranslationRequestCdtInterface
+   *   The translation request.
+   */
+  private function createTranslationRequest(array $data, ?ContentEntityInterface $entity, array $languages): TranslationRequestCdtInterface {
```
drishu: optional parameters should go last

drishu (changes requested):

modules/oe_translation_cdt/tests/src/CdtTranslationTestTrait.php
intelektron: This one is only nullable

drishu: nullable and default value both make the parameter optional, but not worth blocking for, so I'll resolve this

drishu (approved):

PR #324, opened 2024-07-31
EWPP-4585: Separate EU languages on the translation request form.

nagyad:


upchuk:
This doesn't not belong in this component. The concept of EU vs non EU language is from oe_multiligual, which is not a dependency of this component and we will not maintain it here. This needs to be done in EWCMS.

PR #330, opened 2024-10-21
D8AGE-795: Remove language suffix from callback URLs.

intelektron:
### Description

Remove language suffix from callback URLs.

### DIFF
--- a/modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
+++ b/modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Drupal\oe_translation_cdt\Mapper;
 
+use Drupal\Core\Language\Language;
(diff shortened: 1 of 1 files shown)


modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
```diff
@@ -173,11 +175,17 @@ protected function countVolume(int $character_count): float {
   protected function createCallbacks(): CallbackCollection {
     $job_status_url = Url::fromRoute(
       route_name: 'oe_translation_cdt.request_status_callback',
```
drishu, line 179: Lets add a comment explaining why we did it

drishu (changes requested):

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
intelektron: done

drishu (approved):

PR #332, opened 2024-11-15
D8AGE-871: Fix the callback URLs in the translation request.

intelektron:
### Description

Right now, the callbacks in the `modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php:173` file are assigned to wrong routes. The `oe_translation_cdt.request_status_callback` should be assigned to `$request_status_url`

### DIFF
--- a/modules/oe_translation_cdt/modules/oe_translation_cdt_mock/oe_translation_cdt_mock.install
+++ b/modules/oe_translation_cdt/modules/oe_translation_cdt_mock/oe_translation_cdt_mock.install
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * @file
+ * OE Translation CDT Mock install file.
+ */
+
+declare(strict_types=1);
+
+/**
+ * Implements hook_install().
+ */
+function oe_translation_cdt_mock_install($is_syncing) {
+  \Drupal::state()->delete('cdt.token');
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function oe_translation_cdt_mock_uninstall($is_syncing) {
+  \Drupal::state()->delete('cdt.token');
+}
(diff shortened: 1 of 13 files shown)


modules/oe_translation_cdt/modules/oe_translation_cdt_mock/oe_translation_cdt_mock.install
```diff
@@ -0,0 +1,22 @@
+<?php
```
intelektron, line 1: I also added this change to make testing a bit more reliable, previously the mock module would throw an error if the old token was still valid.

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
```diff
@@ -175,14 +175,14 @@ protected function countVolume(int $character_count): float {
   protected function createCallbacks(): CallbackCollection {
     // Set the LANGUAGE_NOT_SPECIFIED to avoid the language suffix in the URL.
     // @see \Drupal\language\Plugin\LanguageNegotiation\LanguageNegotiationUrl::processOutbound()
-    $job_status_url = Url::fromRoute(
+    $request_status_url = Url::fromRoute(
```
drishu, line 178: Cant belive we missed it, lets improve also in TranslationRequestMapperTest the testing of getCallbackBaseUrl

drishu (changes requested):
Not sure since when, but drone seems to fail in cloning the cdt client

modules/oe_translation_cdt/src/Mapper/TranslationRequestMapper.php
intelektron: True! I've just added some additional asserts.

intelektron:
I rebased the epic onto latest 2.x, and fixed the remaining GrumPHP issues. We will be deploying the module with R29 anyway, and by doing this we get rid of Drone problems.

drishu (approved):

PR #333, opened 2024-11-21
EWPP-4928: Allow editors to globally set auto-sync for requests.

22Alexandra:


intelektron:
I'm afraid that this line will force auto_accept always when it is enabled in the ePoetry plugin:

https://github.com/openeuropa/oe_translation/blob/09c96c7fce8b13f27eadbcb8d35c82f5df1282bd/modules/oe_translation_epoetry/src/EventSubscriber/NotificationsSubscriber.php#L369

intelektron (approved):

sergepavle (approved):

PR #339, opened 2025-02-13
EWPP-5161: Use ePoetry 3.x and update to php8.3.

nagyad:

PR #340, opened 2025-02-24
EWPP-5278: Various performance improvements of the active version calculations

upchuk:


composer.json
```diff
@@ -27,7 +27,7 @@
         "drupal/link_description": "^1.0",
         "drupal/metatag": "^1.16",
         "drupal/paragraphs": "^1.13",
-        "drupal/typed_link": "^2.0",
+        "drupal/typed_link": "2.1.1",
```
sergepavle, line 30: Version is locked. Looks like should be mentioned in the `_readme` section.

sergepavle (dismissed):

hernani (dismissed):

composer.json
upchuk: Version will be unlocked once https://github.com/openeuropa/oe_translation/pull/339 is in. Adding readme.

PR #347, opened 2025-04-14
D8AGE-1065: Add support for PHP 8.3 to CDT.

intelektron:
### Description

Add support for PHP 8.3 to CDT.


drishu:
I've merged the 1066 so no need for this change anymore, good to see it pass

PR #349, opened 2025-04-30
EWPP-5476: Etrans integration.

upchuk:
* Introducing the oe_translation_etrans module that provides a new translation plugin that allows users to translate their content using the DGT etrans.
* Refactoring some of the epoetry logic to reuse generically
* Moving the HTMl content formatter to its own module to be used by both.


modules/oe_translation_content_formatter/src/ContentFormatter/ContentFormatterInterface.php
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Return the file content for the translation request.
+   *
+   * @param \Drupal\oe_translation\Entity\TranslationRequestInterface $request
+   *   The request.
+   *
+   * @return \Drupal\Component\Render\MarkupInterface
+   *   String with the file content.
+   */
+  public function export(TranslationRequestInterface $request): MarkupInterface;
```
intelektron: The `HtmlMarkup` returns the markup, but will it be the case for the interface?

I would implement the interface in CDT, where I export to XML, but it will return a string generated by Symfony serializer.

Perhaps `string|MarkupInterface` would make sense.

modules/oe_translation_etrans/modules/oe_translation_etrans_mock/src/Middleware/HttpClientMiddleware.php
```diff
@@ -0,0 +1,63 @@
...
+          // Log the requests done to the etrans mock.
+          $requests = $this->state->get('oe_translation_etrans_mock.mock_requests', []);
+          $requests[] = $request->getBody()->getContents();
+          $request->getBody()->rewind();
+          $this->state->set('oe_translation_etrans_mock.mock_requests', $requests);
+
+          $response = new Response(headers: [
+            'Content-Type' => 'application/json',
+          ], body: json_encode([
+            'requestId' => '55555',
```
intelektron, line 59: I would base this `requestId` on the entity ID somehow, so we don't get a mess when we request multiple translations in the test.

modules/oe_translation_etrans/modules/oe_translation_etrans_mock/src/Middleware/HttpClientMiddleware.php
```diff
@@ -0,0 +1,63 @@
...
+   * Invoked method that returns a promise.
+   */
+  public function __invoke() {
+    return function ($handler) {
+      return function (RequestInterface $request, array $options) use ($handler) {
+        $uri = $request->getUri();
+
+        if (str_contains($uri->getPath(), 'etranslation/api/askTranslate')) {
+          // Log the requests done to the etrans mock.
+          $requests = $this->state->get('oe_translation_etrans_mock.mock_requests', []);
```
intelektron, line 51: Looks like a nice case to put into the `MockLogger`

modules/oe_translation_etrans/src/Controller/EtransController.php
```diff
@@ -0,0 +1,224 @@
...
+  public function error(Request $request) {
+    $content = json_decode($request->getContent());
+    if (!$content || !isset($content->requestId) || !isset($content->externalReference)) {
+      throw new NotFoundHttpException();
+    }
+
+    $remote_id = (string) $content->requestId;
+    $token = (string) $content->externalReference;
+
+    $request = $this->getTranslationRequest($remote_id, $token);
```
intelektron: The var name is confusing, there is already `Request $request`

modules/oe_translation_etrans/src/Controller/EtransController.php
```diff
@@ -0,0 +1,224 @@
...
+    if (!$request) {
+      // We don't care about any failure notification if we cannot determine
+      // a translation request. It's also the only access check we get.
+      $this->logger->warning(sprintf('An etrans delivery was attempted for the request ID %s but for which there was no translation request.', $remote_id));
```
intelektron: Was it eTrans delivery, or just error notification?

modules/oe_translation_etrans/src/Controller/EtransController.php
```diff
@@ -0,0 +1,224 @@
...
+  public function delivery(Request $request) {
+    $content = json_decode($request->getContent());
+    if (!$content || !isset($content->requestId) || !isset($content->externalReference)) {
+      throw new NotFoundHttpException();
+    }
+
+    $remote_id = (string) $content->requestId;
+    $token = (string) $content->externalReference;
+
+    $request = $this->getTranslationRequest($remote_id, $token);
```
intelektron: Same as above

intelektron:
I clicked through translations and didn't find any issues, but I miss the possibility to jump through translation states via the UI, like in ePoetry. I tested it by running the code from tests, which may be a bit unreliable.

modules/oe_translation_etrans/modules/oe_translation_etrans_mock/src/Middleware/HttpClientMiddleware.php
upchuk: It's ok, I made due everywhere with this. Not worth the bother

intelektron (dismissed):

.drone.yml
```diff
@@ -24,6 +24,9 @@ services:
       - EPOETRY_AUTH_CERT_PATH=""
       - EPOETRY_AUTH_CERT_PASSWORD=""
       - EPOETRY_AUTH_EU_LOGIN_BASEPATH=""
+      - ETRANS_SERVICE_URL="https://language-tools.ec.europa.eu/etranslation/api/askTranslate"
```
hernani: Change this value by default to something else

docker-compose.yml
```diff
@@ -25,6 +25,9 @@ services:
       EPOETRY_TICKET_VALIDATION_JOB_ACCOUNT: ""
       EPOETRY_TICKET_VALIDATION_CALLBACK_URL: ""
       EPOETRY_TICKET_VALIDATION_ON: ""
+      ETRANS_SERVICE_URL: "https://language-tools.ec.europa.eu/etranslation/api/askTranslate"
```
hernani: Same

modules/oe_translation_etrans/oe_translation_etrans.install
```diff
@@ -0,0 +1,40 @@
...
+
+    $requirements['etrans.application_name'] = [
+      'title' => t('DGT Etrans application name'),
+      'value' => Settings::get('etrans.application_name', t('Value not set')),
+      'severity' => REQUIREMENT_INFO,
+    ];
+
+    $requirements['etrans.password'] = [
+      'title' => t('DGT Etrans password'),
+      'value' => Settings::get('etrans.password', t('Value not set')),
```
hernani: I wouldn't print this in status report, just mentioned it is set or not set similar to other similar cases.

modules/oe_translation_etrans/src/EtransRequest.php
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /**
+   * Returns the delivery endpoint URL.
+   *
+   * @return string
+   *   The URL.
+   */
+  protected function getDeliveryEndpoint(): string {
+    if (Settings::get('etrans.delivery_endpoint')) {
```
hernani, line 148: Not clear when it is used

upchuk: @hernani This is used for when you wanna override the delivery endpoint with ngrok or something like that for local testing.

modules/oe_translation_etrans/oe_translation_etrans.install
upchuk: Ah true, this was a mistake.

intelektron (dismissed):

intelektron (dismissed):

PR #358, opened 2025-07-15
EWPP-5768: Update to Drupal 10.5.

intelektron:


composer.json
```diff
@@ -8,7 +8,7 @@
     "require": {
         "php": ">=8.1",
         "cweagans/composer-patches": "^1.7 || ^2",
-        "drupal/core": "^10.3 || ^11",
+        "drupal/core": "^10.4 || ^11",
```
nagyad: Here we had to do 10.3 because of the plugin definitions within the module are supported from 10.3, but we don't have to bump the minimum minor versions, since there are no breaking changes.

nagyad (changes requested):

nagyad (dismissed):

nagyad (approved):

22Alexandra (approved):

PR #397, opened 2026-04-07
Always retrieve the correct version field name from configuration

### LINKED ISSUE #396: Version field name is sometimes hardcoded
The version field name is sometimes hardcoded to `version` in the codebase.
This causes the code to break if a custom field name is chosen.
Since the module entity_version has a way to define which field to use as version field for an entity bundle, that value should be respected always.

brummbar:
Fixes #396.


brummbar:
First commit proves the bug.
Second commit adds the fix.
Third commit attempts to avoid the issue of hardcoding the field name to any value in the future.
Fifth commit makes sure the trait is not changed to hardcode back a single value name.

nagyad (approved):

22Alexandra (approved):

PR #407, opened 2026-07-01
EWPP-7024: Skip translation requests with unavailable entity on the Translation Dashboard.

22Alexandra:


modules/oe_translation_local/src/EventSubscriber/TranslationDashboardAlterSubscriber.php
```diff
@@ -123,6 +123,9 @@ public function alterDashboard(ContentTranslationDashboardAlterEvent $event) {
     $rows = [];
     foreach ($translation_requests as $translation_request) {
       $entity = $translation_request->getContentEntity();
+      if (!$entity instanceof ContentEntityInterface) {
+        continue;
```
upchuk, line 127: @22Alexandra This is all we are gonna keep in this PR. We cannot delete all requests for revisions we deleted. We keep track of them in the version log. So let's just guard against the failure if there are leftover old translation requests.

upchuk (changes requested):

upchuk (approved):

fmmribeiro (approved):

PR #410, opened 2026-07-15
EWPP-7012: Allow more elastic access to translation requests.

intelektron:


22Alexandra (dismissed):

bircher (dismissed):

modules/oe_translation_remote/src/Form/RemoteTranslationReviewForm.php
```diff
@@ -326,9 +341,68 @@ public function access(TranslationRequestInterface $oe_translation_request, Acco
...
+
+    $has_permission = $account->hasPermission('accept translation request') || $account->hasPermission('sync translation request');
+    $access = $has_permission
+      ? AccessResult::allowed()->cachePerPermissions()
+      : AccessResult::forbidden('The user is missing the translation permission.')->cachePerPermissions();
+    if ($access->isAllowed()) {
+      return $access;
+    }
+
+    $event = new TranslationRequestCreateAccessEvent($entity, $account, $access, $oe_translation_request->bundle());
```
intelektron: Change into two events

modules/oe_translation_corporate_workflow/src/Access/LocalTranslationAccessCheck.php
```diff
@@ -0,0 +1,59 @@
...
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Routing\Access\AccessInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Drupal\content_moderation\ModerationInformationInterface;
+
+/**
+ * Restricts creating local translation requests based on moderation state.
+ */
+class LocalTranslationAccessCheck implements AccessInterface {
```
intelektron: This is introduced in place of the deprecated event.

22Alexandra (dismissed):

bircher (dismissed):

upchuk (changes requested):
@intelektron The TranslationRequestRouteProvider creates the route for the preview, there we still use the `translate any entity` permission. Dont we need to update this as well?
We also have oe_translation_active_revision that deal with mapping and stuff. Would this feature not be missing for group translators?

Also, we need a ticket to upadte CDT and etrans as well so it's all consistent.

hernani (approved):
Approving for upchuk

sergepavle (approved):

PR #415, opened 2026-07-29
EWPP-7137: HtmlFormatter quiet mode.

upchuk:


modules/oe_translation_content_formatter/src/ContentFormatter/HtmlFormatter.php
```diff
@@ -85,7 +85,10 @@ public function import(string $file, TranslationRequestInterface $request): arra
     $translation_data = [];
 
     $dom = new \DOMDocument();
+    $use_errors = libxml_use_internal_errors(TRUE);
     $dom->loadHTML($file);
+    libxml_clear_errors();
```
intelektron: We manipulate on the global state of DOM here, so if `loadHTML` fails, but if the exception is intercepted, the global state will not be restored. Maybe try..finally would be safer?

```php
try {
    $dom->loadHTML($file);
} finally {
    libxml_clear_errors();
    libxml_use_internal_errors($previous_errors);
}
```

intelektron (approved):

bircher (approved):

PR #418, opened 2026-08-12
Fix crash on translation dashboard when referenced revision no longer exists

### LINKED ISSUE #417: Fix crash on translation dashboard when referenced content entity revision no longer exists
### Problem

When a content entity revision referenced by a local translation request is deleted (e.g. by the `node_revision_delete` module), the translation dashboard crashes with:

```
Error: Call to a member function getRevisionId() on null in
TranslationDashboardAlterSubscriber->alterDashboard()
```

This happens because `TranslationRequestLocal::getContentEntity()` returns `NULL` when `EntityRevisionWithTypeItemList::referencedEntities()` calls `$storage->loadRevision()` on a purged revision. The subscriber does not guard against this case.

Additionally, even without the crash, the orphaned translation request still blocks the "Add new local translation" link in the "Existing translations" table (via `addLocalTranslationOperation()`), leaving the language in an inconsistent state — it appears neither in "Started local translations" nor with an "Add new" operation.

### Root cause

In `TranslationDashboardAlterSubscriber::alterDashboard()`, the `$translation_requests` array is filtered only by status (`!= Synchronised`), then passed directly to `addLocalTranslationOperation()` and iterated to build table rows. If a request's referenced revision has been deleted, `getContentEntity()` returns `NULL`, causing:

1. A fatal error at `$entity->getRevisionId()` (line 138)
2. Even if that were guarded, the language would still be incorrectly blocked from getting a new local translation link

### Proposed fix

Filter out translation requests whose content entity can no longer be loaded, immediately after the status filter and **before** `addLocalTranslationOperation()` is called:

```php
$translation_requests = array_filter($translation_requests, function (TranslationRequestLocal $translation_request) {
  return $translation_request->getTargetLanguageWithStatus()->getStatus() !== TranslationRequestLocal::STATUS_LANGUAGE_SYNCHRONISED;
});
// Filter out translation requests whose referenced content entity revision
// no longer exists (e.g. after revision cleanup/purge by
// node_revision_delete). These orphaned requests should not block the "Add
// new local translation" operation nor appear in the started translations.
$translation_requests = array_filter($translation_requests, function (TranslationRequestLocal $translation_request) {
  return $translation_request->getContentEntity() !== NULL;
});

$this->addLocalTranslationOperation($build, $translation_requests, $current_entity);
```

### Steps to reproduce

1. Create a node with local translation requests
2. Wait for (or trigger) revision cleanup that deletes the revision referenced by an open translation request
3. Visit the node's translation dashboard (`/node/{nid}/translations`)
4. Result: HTTP 500 error

### Environment

- oe_translation version: 3.0.0
- Drupal: 11.4
- Module triggering revision cleanup: `node_revision_delete` 2.1.1

lisotton:
Closes #417.

### DIFF
--- a/modules/oe_translation_local/src/EventSubscriber/TranslationDashboardAlterSubscriber.php
+++ b/modules/oe_translation_local/src/EventSubscriber/TranslationDashboardAlterSubscriber.php
@@ -99,7 +99,10 @@ class TranslationDashboardAlterSubscriber implements EventSubscriberInterface {
     /** @var \Drupal\oe_translation_local\TranslationRequestLocal[] $translation_requests */
     $translation_requests = $storage->getTranslationRequestsForEntity($current_entity, 'local');
     $translation_requests = array_filter($translation_requests, function (TranslationRequestLocal $translation_request) {
-      return $translation_request->getTargetLanguageWithStatus()->getStatus() !== TranslationRequestLocal::STATUS_LANGUAGE_SYNCHRONISED;
+      // Skip synchronised requests and requests whose referenced content
+      // entity revision no longer exists (e.g. after revision cleanup/purge).
+      return $translation_request->getTargetLanguageWithStatus()->getStatus() !== TranslationRequestLocal::STATUS_LANGUAGE_SYNCHRONISED
+        && $translation_request->getContentEntity() !== NULL;
     });
 
     $this->addLocalTranslationOperation($build, $translation_requests, $current_entity);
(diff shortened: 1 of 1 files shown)


upchuk:
@lisotton Can you rebase please?

lisotton:
@upchuk but what should be done here? Remove the change applied in #407 as the fix from this PR is fixing both things?

upchuk:
@lisotton just rebase the PR so i can merge it

lisotton:
@upchuk I rebased the PR, but I also removed what was implemented in #407.

It does not make sense to keep an `if` inside of the loop, if the check is already done inside of the `array_filter`.

